PR #8, opened 2025-07-02
refactor(useKeydown): optimize keydown event handling

AndreyYolkin:
use registrar and single event listener to listen to keydown in a more performant way

### DIFF
--- a/packages/0/src/composables/useKeydown/index.ts
+++ b/packages/0/src/composables/useKeydown/index.ts
@@ -1,44 +1,48 @@
 /**
  * @module useKeydown
  *
+ * @see https://0.vuetifyjs.com/composables/plugins/use-keydown
+ *
  * @remarks
- * Keydown event listener composable with key filtering.
+ * Keydown event listener plugin with key filtering and automatic cleanup.
  *
  * Key features:
- * - Key-specific event handling
+ * - Global keyboard event coordination
+ * - Key-specific event handling with registry
  * - preventDefault and stopPropagation options
  * - Automatic cleanup on scope disposal
- * - Auto-starts when in component scope
+ * - Plugin-first architecture for app-level usage
+ *
+ * Plugin-primary composable for managing document-level keyboard events.
  *
  * @example
- * ```vue
- * <script setup lang="ts">
-  * import { useKeydown } from '@vuetify/v0'
-  *
-  * useKeydown({
-  *   key: 'Enter',
-  *   handler: (event) => {
-  *     console.log('Enter key pressed!', event)
-  *   },
-  *   preventDefault: true,
-  * })
- * </script>
+ * ```ts
+ * import { createApp } from 'vue'
+ * import { createKeydownPlugin } from '@vuetify/v0'
+ *
+ * const app = createApp(App)
+ * app.use(createKeydownPlugin())
  * ```
  */
+
+// Factories
+import { createPlugin } from '#v0/composables/createPlugin'
+import { createTrinity } from '#v0/composables/createTrinity'
+import { createContext, useContext } from '#v0/composables/createContext'
 
 // Composables
 import { useRegistry } from '#v0/composables/useRegistry'
 
 // Utilities
-import { onMounted, getCurrentScope, onScopeDispose, ref, shallowReadonly, shallowRef } from 'vue'
-import { isArray } from '#v0/utilities'
+import { onMounted, getCurrentScope, onScopeDispose, shallowReadonly, shallowRef } from 'vue'
 
 // Globals
 import { IN_DOCUMENT } from '#v0/constants'
 
 // Types
-import type { ID } from '#v0/types'
-import type { RegistryTicket } from '#v0/composables/useRegistry'
+import type { RegistryTicket, RegistryContext, RegistryOptions } from '#v0/composables/useRegistry'
+import type { App, Ref } from 'vue'
+import type { ContextTrinity } from '#v0/composables/createTrinity'
 
 export interface KeyHandler {
   /**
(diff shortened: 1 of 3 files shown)


packages/v0/src/composables/useKeydown/index.ts
```diff
@@ -49,11 +49,13 @@ export function useKeydown (handlers: KeyHandler[] | KeyHandler, options: UseKey
 
   const startListening = () => {
     if (!isListening.value) {
-      handlerIds.value = keyHandlers.map(handler => {
-        const id = crypto.randomUUID()
-        handlerMap.set(id, handler)
-        return id
-      })
+      const ids = Array.from({ length: keyHandlers.length }, () => crypto.randomUUID())
```
userquin: can you remove using crypto on client side? this can force using https for local dev or tests (when not using localhost)

userquin: https://github.com/vitest-dev/vitest/pull/6079

johnleider: Could probably use **useId** if it's not needed before runtime. I do this as well in useRegistrar so I'll need to update its process.

AndreyYolkin: We can't rely on useId here, instead we can make a custom id generator, like this

```ts
export function uuid () {
  if (typeof crypto.randomUUID === 'function') {
    return crypto.randomUUID()
  }

  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, c => {
    const r = Math.trunc(Math.random() * 16)
    const v = c === 'x' ? r : (r & 0x3 | 0x8)
    return v.toString(16)
  })
}

```

userquin: why not? we can force to use vue 3.5 @johnleider

userquin: if so don't use it inside any computed, using getCurrentInstance https://github.com/vuejs/core/blob/eca0e1ccffbee146da0e626f6923715a84355b75/packages/runtime-core/src/helpers/useId.ts#L8

userquin: can you try your suggestion using `HOST=0.0.0.0 vite` opening your ip local address in the browser instead localhost?

AndreyYolkin: We are not callung useId in the composable, but in callback it returns

packages/0/src/composables/useKeydown/index.ts
```diff
@@ -8,27 +9,78 @@ export interface KeyHandler {
...
+      if (handlerMap.size > 0) {
+        startGlobalListener()
+      }
+
+      isListening.value = true
+    }
   }
 
-  if (getCurrentScope()) {
+  const stopListening = () => {
```
johnleider: What is your philosophy behind const functions and when to use them?

AndreyYolkin: I used to write functions via const, but on top level it's converted by antfu eslint plugin automatically, unlike nesting 🥲

packages/0/src/composables/useKeydown/index.ts
```diff
@@ -8,27 +9,78 @@ export interface KeyHandler {
...
+  if (typeof document === 'undefined') return
+
+  if (globalListener && handlerMap.size === 0) {
+    document.removeEventListener('keydown', globalListener)
+    globalListener = null
+  }
+}
+
+export function useKeydown (handlers: KeyHandler[] | KeyHandler, options: UseKeydownOptions = {}) {
+  const { autoStart = true } = options
```
johnleider: How about "eager" or "immediate" instead of "autoStart"?

johnleider (changes requested):
I may be reading this wrong, but I really thing you might benefit from using useRegistry instead of managing handlerMap.

packages/0/src/composables/useKeydown/index.ts
```diff
@@ -8,6 +9,39 @@ export interface KeyHandler {
...
+  immediate?: boolean
+}
+
+let globalListener: ((event: KeyboardEvent) => void) | null = null
+const handlerMap: Map<ID, KeyHandler> = new Map()
+
+function startGlobalListener () {
+  if (typeof document === 'undefined') return
+
+  if (!globalListener) {
```
J-Sek: whenever `if { ... }` guards most of the logic within a function it should be replaced with early exit `if (!..) return`, so we avoid unnecessary indentation and the inner logic is easier to read/scan-through.

There are 3 methods here that would benefit from this quick refactoring.

johnleider:
Looks good but I don't quite understand exporting the handlerMap, waht's the use case?

packages/0/src/composables/useKeydown/index.ts
```diff
@@ -1,77 +1,333 @@
...
+
+export interface KeydownContext<Z extends KeyHandlerTicket = KeyHandlerTicket> extends RegistryContext<Z> {
+  /**
+   * Reactive state indicating whether keyboard listeners are active.
+   */
+  isListening: Readonly<Ref<boolean>>
+  /**
+   * Manually start listening for keyboard events.
+   */
+  startListening: () => void
```
johnleider, line 127: Should probably be named **start** to match other apis in v0

packages/0/src/composables/useKeydown/index.ts
```diff
@@ -1,77 +1,333 @@
...
+  /**
+   * Manually stop listening for keyboard events.
+   */
+  stopListening: () => void
```
johnleider, line 131: Should probably be named **start** to match other apis in v0

packages/0/src/composables/useKeydown/index.ts
```diff
@@ -1,77 +1,333 @@
...
+> (_options: KeydownOptions = {}): E {
+  const { immediate = true, ...options } = _options
+  const registry = useRegistry<Z, E>(options)
+  const isListening = shallowRef(false)
+  let documentListener: ((event: KeyboardEvent) => void) | null = null
+
+  function startDocumentListener () {
+    if (documentListener || !IN_DOCUMENT) return
+
+    documentListener = (event: KeyboardEvent) => {
```
johnleider, line 185: Should this use **useDocumentEventListener**?

johnleider:
Replaced by useHotkey

PR #9, opened 2025-07-05
feat(markdown): add new suite of features

johnleider:


packages/v0/package.json
```diff
@@ -17,6 +17,11 @@
     "vue": ">=3.3.0"
   },
   "devDependencies": {
+    "@types/markdown-it": "^14.1.2",
```
AndreyYolkin: should be listed as peerDependencies instead, see https://www.npmjs.com/package/@vueuse/integrations?activeTab=code package.json

PR #18, opened 2025-07-29
feat(useLayout): add new composable

johnleider:
WIP

### DIFF
--- a/packages/0/src/composables/useLayout/index.ts
+++ b/packages/0/src/composables/useLayout/index.ts
@@ -3,10 +3,12 @@ import { useGroup } from '#v0/composables/useGroup'
 
 // Utilities
 import { computed, shallowReactive, shallowRef, onUnmounted, onMounted, getCurrentInstance } from 'vue'
-import type { ComputedRef, Ref } from 'vue'
+
+// Globals
 import { IN_BROWSER } from '#v0/constants/globals.ts'
 
 // Types
+import type { ComputedRef, ShallowReactive, ShallowRef } from 'vue'
 import type { GroupContext, GroupTicket } from '#v0/composables/useGroup'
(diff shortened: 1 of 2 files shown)


packages/0/src/composables/useLayout/index.ts
```diff
@@ -0,0 +1,125 @@
...
+      id,
+    }
+
+    const ticket = registry.register(item)
+    sizes.set(ticket.id, ticket.size)
+
+    return ticket
+  }
+
+  function unregister (id: ID) {
```
johnleider: Instead of this you could do:

```ts
registry.on('unregister', item => {
  sizes.delete(item.id)
})
```

packages/0/src/composables/useLayout/index.ts
```diff
@@ -0,0 +1,125 @@
...
+    for (const item of registry.collection.values()) {
+      if (item.position === position) {
+        total += sizes.get(item.id) ?? item.size ?? 0
+      }
+    }
+    return total
+  }
+
+  function register (registrant: Partial<Z>): Z {
+    const id = registrant.id ?? genId()
```
johnleider: This is technically being done in useSelection. Since we aren't needing to access the id before registration, you could opt to drop its instantiation and usage on line 96

packages/0/src/composables/useLayout/index.ts
```diff
@@ -0,0 +1,125 @@
+// Composables
+import { useGroup } from '#v0/composables/useGroup'
+
+// Utilities
+import { computed, shallowReactive, shallowRef, type ComputedRef, type Ref, onUnmounted, onMounted, getCurrentInstance } from 'vue'
```
johnleider: separate types out please (working on a way to do this with lint.

PR #19, opened 2025-07-30
feat: add useEventListener composable

AndreyYolkin:
Add new useEventListener composable for handling DOM events. Includes type support for Window, Document, and HTMLElement events, convenience aliases for window and document events

Also add md file to docs and tests

### DIFF
--- a/packages/0/src/composables/useEventListener/index.ts
+++ b/packages/0/src/composables/useEventListener/index.ts
@@ -6,7 +6,7 @@ import { toArray } from '#v0/transformers'
 import type { MaybeArray } from '#v0/types'
 import type { MaybeRef, MaybeRefOrGetter } from 'vue'
 export type CleanupFunction = () => void
-export type EventHandler<E = Event> = (evt: E) => void
+export type EventHandler<E = Event> = (event: E) => void
 
 /**
  * Attaches event listeners to the window object with automatic cleanup.
@@ -22,7 +22,7 @@ export type EventHandler<E = Event> = (evt: E) => void
 export function useEventListener<E extends keyof WindowEventMap> (
   target: Window,
   event: MaybeRefOrGetter<MaybeArray<E>>,
-  listener: MaybeRef<MaybeArray<(this: Window, ev: WindowEventMap[E]) => any>>,
+  listener: MaybeRef<MaybeArray<(this: Window, event: WindowEventMap[E]) => any>>,
   options?: MaybeRefOrGetter<boolean | AddEventListenerOptions>,
 ): CleanupFunction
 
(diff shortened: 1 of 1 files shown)


apps/docs/src/pages/composables/system/use-event-listener.md
```diff
@@ -0,0 +1,301 @@
...
+// Handle keyboard shortcuts globally
+useDocumentEventListener('keydown', (e) => {
+  if (e.ctrlKey && e.key === 's') {
+    e.preventDefault()
+    console.log('Save shortcut triggered')
+  }
+})
+
+// Interactive card with hover effects
+const card = useTemplateRef<HTMLDivElement>()
```
userquin: shouldn't be `const card = useTemplateRef('card')`? `HTMLDivElement` should be inferred.

I **always** use `Ref` suffix => `const cardRef = useTemplateRef('cardRef ')` and `<div ref="cardRef"...>`

packages/0/src/composables/useEventListener/index.ts
```diff
@@ -0,0 +1,170 @@
...
+
+export function useEventListener (
+  target: MaybeRefOrGetter<EventTarget | null | undefined>,
+  event: MaybeRefOrGetter<MaybeArray<string>>,
+  listener: MaybeRef<MaybeArray<EventHandler>>,
+  options?: MaybeRefOrGetter<boolean | AddEventListenerOptions>,
+): CleanupFunction {
+  const cleanups: CleanupFunction[] = []
+
+  const cleanup = () => {
```
johnleider:
```suggestion
  function cleanup () {
```

packages/0/src/composables/useEventListener/index.ts
```diff
@@ -0,0 +1,170 @@
...
+      for (const evt of eventList) {
+        for (const listenerFn of listenerList) {
+          cleanups.push(register(el, evt, listenerFn, opts))
+        }
+      }
+    },
+    { immediate: true, flush: 'post' },
+  )
+
+  const stop = () => {
```
johnleider:
```suggestion
  function stop () {
```

packages/0/src/composables/useEventListener/index.ts
```diff
@@ -0,0 +1,170 @@
...
+ * @param target Window object to attach listeners to.
+ * @param event Event name(s) to listen for from WindowEventMap.
+ * @param listener Event handler function(s) with proper window event typing.
+ * @param options Optional event listener configuration.
+ * @returns Function to manually remove all attached listeners.
+ */
+export function useEventListener<E extends keyof WindowEventMap> (
+  target: Window,
+  event: MaybeRefOrGetter<MaybeArray<E>>,
+  listener: MaybeRef<MaybeArray<(this: Window, ev: WindowEventMap[E]) => any>>,
```
johnleider: There's a mix of variable names for events. I think just play *event* is fine.

johnleider (changes requested):
I'm not going to leave feedback on all of them, but the preference for functions would be to use regular instead of const. Lookin really good.

PR #23, opened 2025-08-04
refactor: useLayout plugin and dynamic sizing

Haviles04:
refactored layout to use as a plugin


packages/0/src/composables/useLayout/index.ts
```diff
@@ -18,14 +29,15 @@ export interface LayoutTicket extends GroupTicket {
...
 export interface LayoutContext<Z extends LayoutTicket> extends GroupContext<Z> {
   bounds: {
-    top: ComputedRef<number>
-    bottom: ComputedRef<number>
-    left: ComputedRef<number>
-    right: ComputedRef<number>
+    top: Ref<number>
+    bottom: Ref<number>
+    left: Ref<number>
+    right: Ref<number>
```
johnleider: Should be ShallowRef

packages/0/src/composables/useLayout/index.ts
```diff
@@ -36,25 +48,54 @@ export interface LayoutContext<Z extends LayoutTicket> extends GroupContext<Z> {
...
+  left: ShallowRef<number>
+  right: ShallowRef<number>
+  top: ShallowRef<number>
+  bottom: ShallowRef<number>
+  resize: () => void
+  register: (item?: Partial<Z>) => Z
+}
+
+export interface LayoutOptions extends GroupOptions {
+  el?: Ref<HTMLElement | null>
```
johnleider: Should be ShallowRef

packages/0/src/composables/useAtomRef/index.ts
```diff
@@ -0,0 +1,24 @@
+// Utilities
```
johnleider: We need to check the performance implications of needing to add this layer.

johnleider (changes requested):

johnleider (approved):
Looks great! Only feedback is should we be using our built in resize observer?

packages/0/src/composables/useLayout/index.ts
```diff
@@ -63,64 +126,196 @@ export function useLayout<
...
       width.value = window.innerWidth
+      height.value = window.innerHeight
+      left.value = 0
+      top.value = 0
+      right.value = window.innerWidth
+      bottom.value = window.innerHeight
     }
+  }
 
+  let observer: ResizeObserver | null = null
```
userquin: `let observer: ResizeObserver | undefined`

apps/docs/src/components/app/AppBanner.vue
```diff
@@ -1,25 +1,41 @@
...
+  // Utilities
+  import { useTemplateRef } from 'vue'
 
   // Types
   import type { AtomProps } from '@vuetify/v0'
 
   const { as = 'header' } = defineProps<AtomProps>()
 
   const breakpoints = useBreakpoints()
+  const element = useTemplateRef<HTMLElement>('banner')
```
userquin: why `<HTMLElement>`? this should be resolved to atom instance by vue

packages/0/src/composables/useLayout/index.ts
```diff
@@ -63,64 +126,196 @@ export function useLayout<
...
+
+  if (IN_BROWSER && getCurrentInstance()) {
```
userquin, line 240: still using `IN_BROWSER`? this can break Nuxt SSR, we'll need to add some "resolver" for this

apps/docs/src/components/app/AppBanner.vue
Haviles04: It's the only way I could work around TS yelling about the `size` computed in the register function. Since defineExpose forces you to wrap the element in an object it causes a type mismatch with useTemplateRef in the same component vs calling it on a child

Haviles04: very much open to solutions on this

userquin: `element.value?.$el.size`?

userquin: Or maybe use defineExpose

userquin: Uhmm, will check atom...

userquin: `element.value?.$refs.element.size`?

Haviles04: I think it's more of a TS issue between defineExpose 

this is the error I get when I remove `<HTMLElement>`

```
Readonly<ShallowRef<ShallowUnwrapRef<{
    element: Readonly<ShallowRef<HTMLElement | null>>;
}> | null>>
is not assignable to type ElementSource | undefined
Type
Readonly<ShallowRef<ShallowUnwrapRef<{
    element: Readonly<ShallowRef<HTMLElement | null>>;
}> | null>>
is not assignable to type
ShallowRef<Record<string, ShallowRef<HTMLElement | null>> | null>
```


Also TS says value doesn't exist on ShallowUnwrapRef

userquin: how about adding `  | ShallowRef<DOMElement | null>` to the `ElementSource`?
```ts
export type ElementSource =
  | ShallowRef<HTMLElement | null>
  | ShallowRef<DOMElement | null>
  | ShallowRef<Record<string, ShallowRef<HTMLElement | null>> | null>
```
or just replace `| ShallowRef<HTMLElement | null>` with `| ShallowRef<DOMElement | null>` (if so replace also the shallowref record type)

Haviles04: ShallowRef<DOMElement | null> was it! THANK YOU

packages/0/src/composables/useLayout/index.ts
Haviles04: if we just ran this onMounted would we run into weird hydration issues?

userquin: We can, since we have suspense and the transition api, we need to use the `app:suspense:resolve`: I use a pinia store + isReady (hydrated here), then on the hook we just update the pinia flag.

At docs app, we can use pinia + onMounted at default layout.

I guess maybe we don't need this hack if we can move the css in js to just css via some cssvar => the problem with the nav drawer (CLS) is about the `useBreakpoint` => indirect css in js (maybe we can move that code to the theme restoration script)

I'm going to play on my local trying to inline css styles and some vite-ssg build hooks + beasties

PR #25, opened 2025-08-06
docs: add component pages

MajesticPotatoe:
Adds missing component pages

## ToDo
- [ ] Needs to be checked for accuracy


johnleider:
I've held off on these simply because they are based off of composables which have been in flux. We can use this as an opportunity to explore page structure so we can maybe put together a few .md templates that solve the scenarios:

- I want to add a new component, what docs does it need?
- I want to add a new composable, what does does it need?

MajesticPotatoe:
> I've held off on these simply because they are based off of composables which have been in flux. We can use this as an opportunity to explore page structure so we can maybe put together a few .md templates that solve the scenarios:

All good, i just saw bare pages, and thought, well get something on it at least.

My thoughts on minimum component structure:
```
#Component
Description / details / structure breakdown

## API
## Examples
Interactive Usage Example (like we got in main docs maybe, that showcases the props/slots/events + output)

-- examples should outline at min, one fore each prop/slot/event group as applicable)
### Props
### Slots
### Events
```

Composable pretty much follow a simpler structure (at least as a starting point)

```
#Composable
Description / details

## API
## Examples
-- examples of each exported function and its use
```

johnleider:
Add # Usage and I think that's good.

PR #27, opened 2025-08-08
docs(use-registry): update page content

antonhansson0624:


apps/docs/src/pages/composables/registration/use-registry.md
```diff
@@ -127,3 +127,282 @@ console.log(registry.size) // 3
...
+  
+### `browse`
+
+- **Type**
+  ```ts
+  function browse(value: unknown): ID | ID[] | undefined
+  ```
+
+- **Details**
+  Searches for item(s) by their value and returns the corresponding ID(s), or undefined if none match.
```
johnleider: Add info that says like values will be pushed into an array. So if the user has 2 items with the same value, calling browse for that value will return an array of IDs.

apps/docs/src/pages/composables/registration/use-registry.md
```diff
@@ -127,3 +127,282 @@ console.log(registry.size) // 3
...
+  const registry = useRegistry()
+  registry.register({ id: 'apple', value: 'Apple' })
+  registry.register({ id: 'banana', value: 'Banana' })
+
+  console.log(registry.lookup(0)) // 'apple'
+  console.log(registry.lookup(1)) // 'banana'
+  console.log(registry.lookup(5)) // undefined
+  ```
+
+### `values`
```
johnleider, line 250: values, entries, and keys are cached for performance. Can add something about that.

antonhansson0624: I will add sth into Details.

apps/docs/src/pages/composables/registration/use-registry.md
antonhansson0624: I will add certainly.

johnleider (approved):

PR #32, opened 2025-08-16
feat: maintenance global app state

antonhansson0624:


johnleider:
Are these changes valid?

PR #38, opened 2025-09-02
docs: migrate to vite-ssg for static site generation

AndreyYolkin:
- Add vite-ssg configuration and update build script
- Implement SSR support with pinia state management
- Move router logic to main.ts and handle dynamic import errors
- Update AppBar.vue to support SSR

### DIFF
--- a/apps/docs/src/main.ts
+++ b/apps/docs/src/main.ts
@@ -20,7 +20,7 @@ export const createApp = ViteSSG(
       pinia.state.value = initialState.pinia || {}
 
     if (!import.meta.env.SSR) {
-    // Workaround for https://github.com/vitejs/vite/issues/11804
+      // Workaround for https://github.com/vitejs/vite/issues/11804
       router.onError((err, to) => {
         if (err?.message?.includes?.('Failed to fetch dynamically imported module')) {
           if (localStorage.getItem('vuetify:dynamic-reload')) {


apps/docs/package.json
```diff
@@ -23,7 +23,9 @@
     "unplugin-vue-markdown": "^29.1.0",
     "unplugin-vue-router": "^0.14.0",
     "vite": "catalog:",
-    "vite-plugin-vue-layouts-next": "^1.0.0"
+    "vite-plugin-vue-layouts-next": "^1.0.0",
+    "vite-ssg": "^28.1.0",
+    "vue-router": "^4.5.1"
```
johnleider, line 28: Should either of these be in the catalog? What is the criteria for that?

apps/docs/src/main.ts
```diff
@@ -1,11 +1,43 @@
...
+    app.use(pinia)
 
-app.mount('#app')
+    if (import.meta.env.SSR)
+      initialState.pinia = pinia.state.value
+    else
+      pinia.state.value = initialState.pinia || {}
+
+    if (!import.meta.env.SSR) {
+    // Workaround for https://github.com/vitejs/vite/issues/11804
```
johnleider:
```suggestion
      // Workaround for https://github.com/vitejs/vite/issues/11804
```

apps/docs/package.json
AndreyYolkin: We are not using these packages anywhere besides docs. We can implement another policy and put every dependency we have at catalog

johnleider: Makes sense, if it's only in 1 project. I'm okay with either policy direction.

johnleider (approved):

PR #40, opened 2025-09-10
feat: new composable useTimeline

Haviles04:
new composable built on top of registry, with ability to undo/redo last actions

### DIFF
--- a/packages/0/src/composables/useTimeline/index.ts
+++ b/packages/0/src/composables/useTimeline/index.ts
@@ -1,43 +1,52 @@
-import { useRegistry } from '#v0'
-import type { RegistryContext, RegistryTicket } from '#v0'
+// Composables
+import { useRegistry } from '#v0/composables/useRegistry'
+
+// Types
+import type { RegistryContext, RegistryOptions, RegistryTicket } from '#v0/composables/useRegistry'
 
 export interface TimelineContext<Z extends TimelineTicket> extends RegistryContext<Z> {
-  size: number
+  /* Unapplies the last registered ticket */
   undo: () => void
+  /* Reapplies the last undone ticket */
   redo: () => void
 }
 
-export interface TimelineTicket extends RegistryTicket {
-  id: string
-  index: number
-  value: unknown
-}
+export interface TimelineTicket extends RegistryTicket {}
 
-export interface TimelineOptions {
+export interface TimelineOptions extends RegistryOptions {
   size?: number
 }
 
-export function useTimeline<Z extends TimelineTicket = TimelineTicket,
-  E extends TimelineContext<Z> = TimelineContext<Z>> (_options: TimelineOptions) {
+/**
+ * Creates a registry with timeline capabilities (undo/redo)
+ *
+ * @param _options Optional configuration for timeline
+ * @template Z The type of ticket to be stored in the timeline
+ * @template E The type of the timeline context
+ * @returns The timeline context object
+ *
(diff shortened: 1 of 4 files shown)


packages/0/src/composables/useHistory/index.ts
```diff
@@ -0,0 +1,93 @@
...
+  index: number
+  value: unknown
+}
+
+export function useHistory<Z extends HistoryTicket = HistoryTicket,
+  E extends HistoryContext<Z> = HistoryContext<Z>>
+(_options: HistoryOptions = {}) {
+  const {
+    size = 10,
+  } = _options
```
johnleider: I typically only use _options if I'm going to destructure into an options object.

packages/0/src/composables/useHistory/index.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    restore()
+  }
+
+  function restore () {
+    const value = firstOutValues.pop()
+
+    const fullArray = value ? [value, ...registry.values()] : [...registry.values()]
+    registry.clear()
+    for (const item of fullArray) {
```
johnleider: Could do registry.onboard

johnleider (changes requested):

packages/0/src/composables/useTimeline/index.ts
```diff
@@ -0,0 +1,82 @@
+import { useRegistry } from '#v0'
+import type { RegistryContext, RegistryTicket } from '#v0'
+
+export interface TimelineContext<Z extends TimelineTicket> extends RegistryContext<Z> {
+  timeline: TimelineTicket[]
+  size: number
+  register: (item?: Partial<Z>) => Z
```
johnleider: Shouldn't need to be redefined.

packages/0/src/composables/useTimeline/index.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+export interface TimelineTicket extends RegistryTicket {
+  id: string
+  index: number
+  value: unknown
+}
+
+export function useTimeline<Z extends TimelineTicket = TimelineTicket,
+  E extends TimelineContext<Z> = TimelineContext<Z>> (size = 10) {
```
johnleider: To make it easier to pivot in the future, I would suggest moving `size` into an `options` object:

```ts
const timeline = useTimeline({ size: 10 })
```

This also allows the turning on of events:

```ts
const timeline = useTimeline({ size: 5, events: true })
```

packages/0/src/composables/useTimeline/index.ts
```diff
@@ -0,0 +1,82 @@
...
+  const registry = useRegistry<Z, E>()
```
johnleider:
```ts
const { size, ...options } = _options

const registry = useRegistry<Z, E>(options)
```

packages/0/src/composables/useTimeline/index.ts
```diff
@@ -0,0 +1,82 @@
+import { useRegistry } from '#v0'
+import type { RegistryContext, RegistryTicket } from '#v0'
+
+export interface TimelineContext<Z extends TimelineTicket> extends RegistryContext<Z> {
+  timeline: TimelineTicket[]
```
johnleider: For now I'm wanting to leave any alias esque values out until we have a meeting on what those look like. For now, the docs should tell the user they can access the timeline by calling `timeline.values()`

PR #42, opened 2025-09-10
build: simplify package exports configuration

AndreyYolkin:
Use https://tsdown.dev/options/package-exports#conditional-dev-exports to let tsdown generate exports automatically

Note about ./browser: it's needed in some extra rare cases, because systems like jsdelivr provide version with prebundled env variables, which is enough for Vuetify Play to run. We don't need to have an additional dts file then


johnleider:
I just had to make a bunch of changes to our build process to get Vuetify0 to work in an app outside of the documentation. This probably caused your merge conflicts, sorry.

johnleider (approved):

PR #47, opened 2025-10-08
feat: Doc block examples

Haviles04:


packages/0/src/composables/useBreakpoints/index.ts
```diff
@@ -216,16 +229,19 @@ export function createBreakpoints<
...
+ *
+ *   const { isMobile, mdAndUp } = useBreakpoints()
  * </script>
  *
  * <template>
  *   <div class="pa-4">
- *     <p v-if="isMobile">Mobile layout active</p>
- *     <p v-else-if="mdAndUp">Medium and up layout active</p>
+ *     <p v-if="isMobile.value">Mobile layout active</p>
+ *     <p v-else-if="mdAndUp.value">Medium and up layout active</p>
```
Haviles04, line 244: Wont vue figure out the value in the template, since these are just shallowrefs?

johnleider: Unfortunately no

PR #71, opened 2025-12-30
refactor(vue): replace getCurrentInstance with utility functions

AndreyYolkin:
Vue 3.6 comes wtih a different util for getting a generic component instance, which results in getCurrentInstance being nullish for мapor components (see https://github.com/vuejs/core/releases/tag/v3.6.0-beta.1)

Since the codebase doesn't have a unified way to check existing instance without unneeded warnings on dev (such as [getContext](https://github.com/vuejs/core/blob/aac7e1898907445c8f89b22047a9bfcf0a6e91b8/packages/runtime-core/src/apiSetupHelpers.ts#L403)) we need to have a wrapper for that
This PR introduced 2 internal methods to get check instance and get its name universally (tested on 3.5.24 and 3.6.0-beta.1)

### DIFF
--- a/packages/paper/src/utilities/getCurrentInstanceName.ts
+++ b/packages/paper/src/utilities/getCurrentInstanceName.ts
@@ -1,5 +1,7 @@
+// Framework
+import { instanceName } from '@vuetify/v0/utilities'
+
 // Utilities
-import { instanceName } from '#v0/utilities'
 import { toKebabCase } from './helpers'
 
 export function getCurrentInstanceName () {


packages/paper/src/utilities/getCurrentInstanceName.ts
```diff
@@ -1,9 +1,9 @@
...
+import { instanceName } from '#v0/utilities'
 import { toKebabCase } from './helpers'
-import { getCurrentInstance } from 'vue'
 
 export function getCurrentInstanceName () {
-  const vm = getCurrentInstance()?.type
-
-  return toKebabCase(vm?.name?.replace('V0', '') ?? '')
+  const name = instanceName()
+  if (!name) return ''
```
AndreyYolkin, line 9: this early return could save a lot of unneeded `toKebabCase` calls. Tried with 50% `undefined` names, received 50% ops/sec improvement

PR #155, opened 2026-03-12
refactor(useHotkey): use AST for the hotkey string parser

MatthewAry:


packages/0/src/composables/useHotkey/index.ts
```diff
@@ -217,11 +217,17 @@ export function useHotkey (
...
     if (!currentKeys) return
 
-    const groups = splitKeySequence(currentKeys.toLowerCase())
-    if (groups.length === 0) return
+    const parsed = parseKeyCombination(currentKeys.toLowerCase())
+    if (!parsed) return
 
-    isSequence = groups.length > 1
-    keyGroups = groups
+    const parts: Exclude<KeyCombination, Sequence>[] = typeof parsed !== 'string' && parsed.type === 'sequence'
```
johnleider:
```suggestion
    const parts: Exclude<KeyCombination, Sequence>[] = !isString(parsed) && parsed.type === 'sequence'
```

`import { isString } from '#v0/utilites'`

packages/0/src/composables/useHotkey/index.ts
```diff
@@ -284,17 +290,14 @@ function isModifier (key: string): key is Modifier {
...
-  ) as Record<Modifier, boolean>
+const emptyModifiers = Object.fromEntries(
+  MODIFIERS.map(m => [m, false]),
+) as Record<Modifier, boolean>
 
-  if (parts.length === 0) {
-    return { modifiers, actualKey: undefined }
-  }
+function parseKeyGroup (group: Combo | Key): ParsedKeyGroup {
+  const parts = typeof group === 'string' ? [group] : group.parts
```
johnleider:
```suggestion
  const parts = isString(group) ? [group] : group.parts
```

packages/0/src/composables/useHotkey/index.ts
```diff
@@ -308,7 +311,11 @@ function parseKeyGroup (group: string): ParsedKeyGroup {
   return { modifiers, actualKey }
 }
 
-function matchesKeyGroup (e: KeyboardEvent, group: string, platformIsMac: boolean): boolean {
+function matchesKeyGroup (e: KeyboardEvent, group: Exclude<KeyCombination, Sequence>, platformIsMac: boolean): boolean {
+  if (typeof group !== 'string' && group.type === 'alternate') {
```
johnleider:
```suggestion
  if (!isString(group) && group.type === 'alternate') {
```

packages/0/src/composables/useHotkey/parsing.ts
```diff
@@ -12,193 +12,128 @@ export const MODIFIERS = ['ctrl', 'shift', 'alt', 'meta', 'cmd'] as const
...
 
-  return collapsed
+  // key = /./ *(/[^-/+_ ]/)
+  function parseKey (): Key {
+    const ch = peek()
+    if (ch == null) {
+      throw new ParseError('Unexpected end of input')
+    }
+    const next = peek(1)
+    if (isSep(ch) && next != null && !isSep(next)) {
```
johnleider:
```suggestion
    if (isSep(ch) && !isNullOrUndefined(next) && !isSep(next)) {
```

packages/0/src/composables/useHotkey/index.ts
KaelWD: If you really want to enforce this for some reason it should be an eslint rule.

PR #204, opened 2026-04-22
docs: use native api link

ikushum:
## Problem
Was not able to open API doc page in new tab (native browser way).

## Implemented solution
Used `<router-link>` instead of programatic navigation using `router.push`


J-Sek (approved):
Looks good to me.
I still cannot use [Enter] on links but it turned out to be a separate issue (#205)

johnleider (approved):

PR #207, opened 2026-04-23
refactor(createEnrollable): internal factory for dual-mode toggleable roots

johnleider:
## Summary

- Extracts the ~85% structural overlap between `CheckboxRoot` and `SwitchRoot` into an **internal** composable. Not a public primitive.
- Matches the existing `createObserver` / `createFocusTraversal` convention: lives under `packages/0/src/composables/createEnrollable/` but is **not** re-exported from the barrel, **not** documented, and flagged `category: "internal"` in `maturity.json`.
- Consumers import via deep path: `#v0/composables/createEnrollable`.
- Roadmap audit confirmed only 3 components fit the dual-mode binary-state + optional-parent-enrollment shape (Checkbox, Switch, Toggle); nothing on the 27-component roadmap substantially broadens the reuse pool — hence internal.

## Consumers in this PR

- `CheckboxRoot` — migrated (reference consumer)
- `SwitchRoot` — migrated (reference consumer)
- `ToggleRoot` — deferred; needs `isPressed` / `.selection` accessor aliasing to fit the shape
- `RadioRoot` — out of scope (mandatory group + roving tabindex, no v-model)

## Design notes

- Framework-neutral: no DOM, no DI, no lifecycle. Callers resolve the optional parent group via `useXGroup` + try/catch and pass `GroupContext | null` in.
- Vue 3.5 reactive-destructure quirk: consumers wrap props as getters (`disabled: () => toValue(disabled)`) at the call site to preserve prop-change reactivity across the composable boundary.

### DIFF
--- a/packages/0/src/maturity.json
+++ b/packages/0/src/maturity.json
@@ -272,6 +272,11 @@
       "category": "internal",
       "notes": "Internal factory for useRovingFocus and useVirtualFocus"
     },
+    "createEnrollable": {
+      "level": "draft",
+      "category": "internal",
+      "notes": "Internal factory for dual-mode toggleable roots (Checkbox, Switch, Toggle)"
+    },
     "createValidation": {
       "level": "preview",
       "since": "0.1.8",
--- a/packages/0/src/components/Checkbox/CheckboxRoot.vue
+++ b/packages/0/src/components/Checkbox/CheckboxRoot.vue
@@ -18,6 +18,7 @@
 
   // Composables
   import { createContext } from '#v0/composables/createContext'
+  import { createEnrollable } from '#v0/composables/createEnrollable'
 
   // Types
   import type { AtomProps } from '#v0/components/Atom'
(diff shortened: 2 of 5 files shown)


johnleider:
## Discussion: novel internal-composable axis

Flagging this for reviewer input — worth a short discussion before merge.

The existing `"internal"` category covers **composable-consumed factories**:

- `createObserver` → `useResizeObserver`, `useIntersectionObserver`, `useMutationObserver`
- `createFocusTraversal` → `useRovingFocus`, `useVirtualFocus`

Both consolidate cross-cutting logic shared across a family of sibling **composables**. Framework-neutral, no Vue setup semantics.

`createEnrollable` is a different beast — it's consumed by `.vue` component setup blocks (CheckboxRoot, SwitchRoot, and future ToggleRoot), not by other composables. That means it sits adjacent to Vue-specific surface: `defineModel` refs, prop-destructure reactivity, `onBeforeUnmount` lifecycle. The reactive-destructure getter trick at each call site (`disabled: () => toValue(disabled)`) is a concrete artifact of this — it's only a concern because setup is the call site.

### The axis

- **Composable-consumed internals** = dedup within the composable layer
- **Component-consumed internals** = dedup within the component layer — effectively a component base class in disguise (cf. Vuetify's SelectionControl at the UI layer)

This is the first component-consumed internal in v0. If a second shows up (e.g., a shared floating-layer factory for Popover/Dialog/Tooltip roots, or a shared close-button primitive across AlertDialogClose/DialogClose/SnackbarClose), the `"internal"` category starts conflating two axes.

### Options on the table

**A.** Keep `"internal"`, let `notes` carry the axis — current state.

**B.** Split into `"internal"` vs `"internal-component"` — cleanest signal but first-mover forces a convention decision today.

**C.** Add a new schema field (e.g., `"layer": "composable" | "component"`) — backwards-compatible, richer, but schema churn for one entry.

**D.** Defer the split. Keep `"internal"` for now, revisit when a second component-layer internal lands.

My lean is D — one instance isn't enough to justify a schema or category split. But flagging here so we don't silently set precedent without thinking about it.

Also removed the `since` field from the maturity entry in the latest push — for internal factories with no public observability, tracking when they became available is noise. If that aligns with the direction, it's worth doing the same cleanup on `createObserver` / `createFocusTraversal` separately.

Thoughts?

packages/0/src/maturity.json
```diff
@@ -272,6 +272,11 @@
       "category": "internal",
       "notes": "Internal factory for useRovingFocus and useVirtualFocus"
     },
+    "createEnrollable": {
+      "level": "draft",
+      "category": "internal",
+      "notes": "Internal factory for dual-mode toggleable roots (Checkbox, Switch, Toggle)"
```
J-Sek, line 278: `dual-state`? There are technically 3 states, but it does not matter. The `mode` might suggest different behaviors.

Edit: it is possible I misunderstand something. The description above says `dual-mode binary-state`. If it is not just wordy spit from LLM, maybe the mode means something and I am just confused because of dual/binary.

J-Sek:
Also leaning to D. I get we want to have guardrails sooner than later, but leaving a few less-important architectural decisions for later should help us adapt when we see v0 getting adopted.

KaelWD:
I don't know enough about the architecture of this project to have an opinion.

PR #230, opened 2026-05-04
feat(Alert): add new component

J-Sek:
(just a quick draft from a single prompt)

- (docs) the new page does a good job of introducing the component, but we might want to stress out that alerts are not always a good choice for visual callouts scattered across the article.. not sure if it matters - need to test with actual screen readers
- more examples under consideration
  - with actions [approve/cancel] or [read more]
  - with image instead of icon


johnleider:
Closing in favor of #823

PR #241, opened 2026-05-20
chore(builder): complete framework-creation flow

johnleider:
## Summary

Builds `apps/builder/` into a complete car-configurator wizard for assembling a v0-based Vue app from selectable plugins and components — now current with v0 master and able to produce a working framework start to finish.

**The flow:** landing → select plugins (15, incl. new ReducedMotion + Tooltip) → per-plugin config screens with a live preview aside (every configurable plugin has a dedicated preview module) → component picker (maturity-aware: draft components are visibly gated) → review build sheet → **outputs**: a zip starter that installs and builds green, and a playground URL carrying the full plugin configuration.

## What's in this update

- **Merged 429 commits of master drift** and realigned every config screen with current v0 option shapes (permissions tuples, locale without the non-public adapter, rules alias names, top-level theme `foreground`).
- **Codegen correctness**: adapter imports from their real subpaths, third-party adapters emit wiring instructions instead of uncompilable zero-arg constructions, real rule predicates for known aliases (compile-time coverage via `Record<KnownAlias, string>`), explicit plugin install order with storage auto-included when anything persists, undefined config values survive, draft components/composables filtered from emission.
- **Generated starter actually verified**: `pnpm install && pnpm run build` green, styled via an emitted uno.config mapping theme tokens to `var(--v0-*)` with fallbacks, demos rewritten on v0 compounds (zero native form controls).
- **Playground parity**: payload emits `src/main.ts` so the user's configuration installs in the sandbox; `framework.ts` binds through a namespace (vue-repl cannot resolve re-exported bindings).
- **Live preview modules** for all configurable plugins (notifications, logger, storage, stack, features, permissions, date, rules, tooltip, reduced-motion + the original four).
- **A11y**: cards are `Toggle.Root` with real pressed state, every field labeled, CTAs announce their visible text, SVG indicators only.
- **Design pass**: one visual language (graphite + teal, real elevation steps, mono for identifiers), landing hero built from the actual generated `main.ts`, segmented wizard progress, build-sheet review page, dark mode persisted, AA contrast verified.
- **Tests**: `apps/builder` wired into the vitest workspace — 153 tests including drift guards (every config screen must have a factory emission; every known alias must have a predicate).

## Verification

Three full Playwright walkthroughs (desktop/tablet/mobile, light/dark): complete flow, zero console errors, zip built and served, playground rendered with configured theme, state persistence across reload.

## Known limitations

- NumberField commits on blur/Enter (v0 has no eager-commit option), so number fields update the preview on commit rather than per keystroke.
- The playground supplies its own `uno.config.ts`, so a deliberately partial palette can render unstyled borders there (zip output has fallbacks).


johnleider:
Since the last push: **saved builds** (auto-saved multi-build persistence behind an API-swappable seam in `stores/persistence.ts`, landing resume list + header switcher, legacy-key migration), plus a verified second-model review cycle (Codex → adversarial verification): fixed a reproducible build-switch data-loss race with a request-token guard, added an isSaving/saveError contract (load-bearing once the API backend lands), rev-enveloped index writes, output-button error feedback with popup-blocker fallback, delete confirmation via AlertDialog, and five mutation-proven registry/demo invariant tests. 179 tests.

johnleider:
Second-model + multi-agent audit cycle complete. Codex (verified: 9 confirmed/3 rejected) drove the request-token switch guard, persistence error contract, rev-enveloped index and output-button feedback. Three Opus auditors (UX / design / flexibility) drove: saved-builds hardening (reactive cross-tab list, lazy create, honest delete-to-zero), an honest dependency graph (31 phantom edges removed — review sheet no longer contradicts the starter — plus 2 real edges the single-level scan missed), five new registry drift guards (191 tests), and a measured design wave: zero 390px overflow on every route, unified stepper/field treatment, working dark-mode elevation (the dark: variant matched nothing before), warning/on-error tokens at verified AA, sanctioned button ranks. All findings adversarially verified before fixing; all fixes browser-verified before commit.

apps/builder/src/engine/manifest.ts
```diff
@@ -0,0 +1,1620 @@
...
+// alias NAMES; each becomes a real predicate so the emitted object satisfies
+// RulesOptions['aliases'].
+function formatAliases (names: unknown): string | null {
+  if (!Array.isArray(names) || names.length === 0) return null
+
+  const lines = names
+    .filter((name): name is string => typeof name === 'string')
+    .map(name => {
+      const body = isAlias(name) ? ALIASES[name] : `(v: unknown) => true /* TODO: implement */`
+      return `    ${name}: ${body},`
```
johnleider: **[bug]** `formatAliases` interpolates user-supplied alias names as unquoted JS identifiers. RulesConfig.vue accepts any trimmed string (placeholder is `phone`, but kebab-case is the natural form). A name like `foo-bar` or `2fa` emits `foo-bar: (v: unknown) => true` and the generated `src/main.ts` is a syntax error. The same path is an injection sink: `foo}\nprocess.exit(1)//` is written straight into the zip/playground sources. Known aliases happen to be valid identifiers, so the existing tests never catch this.

**Suggestion:** Quote keys with `JSON.stringify(name)` (or reject names that are not `/^[A-Za-z_$][\w$]*$/` in the form and the emitter). Do not emit a stub for a name that is not a valid identifier.

apps/builder/src/engine/manifest.ts
```diff
@@ -0,0 +1,1620 @@
...
+  // boots the same plugin set the starter does. Without a theme the demos have
+  // no palette to resolve against, so preview-only defaults stand in.
+  const themed = plugins.includes('useTheme')
+    ? { plugins, config }
+    : { plugins: [...plugins, 'useTheme'], config: { ...config, useTheme: PREVIEW_THEME } }
+
+  return {
+    files: {
+      ...generateFiles(input),
+      'src/main.ts': generateMainTs(themed.plugins, themed.config, './uno.config.ts'),
```
johnleider: **[bug]** `toHashData` writes `src/main.ts` via `generateMainTs` but never adds the custom date adapter stub. `generateMainTs` with `{ useDate: { adapter: 'custom' } }` imports `CustomDateAdapter` from `./adapters/date`. `zip.ts` `generateSources` adds that file; the playground payload does not. Review.vue passes `pluginConfig` through, so “Open in Playground” after picking the custom adapter 404s the module.

**Suggestion:** Mirror zip.ts: if `needsDateStub(themed.plugins, themed.config)`, set `files['src/adapters/date.ts'] = generateDateStub()`. Add a `toHashData` test that asserts the stub is present when `adapter: 'custom'`.

apps/builder/src/engine/manifest.ts
```diff
@@ -0,0 +1,1620 @@
...
+  useClickOutside: 'disclosure',
+  usePopover: 'disclosure',
+  Dialog: 'disclosure',
+  Popover: 'disclosure',
+  // Observers
+  useResizeObserver: 'observers',
+  useIntersectionObserver: 'observers',
+}
+
+export function generateImports (): Record<string, string> {
```
johnleider: **[bug]** Playground import maps are exact specifiers. `generateImports` only maps `@vuetify/v0` → jsDelivr `dist/index.mjs` and `@vue/devtools-api`. Codegen for Date emits `from '@vuetify/v0/date'` and the adapter statically imports `@js-temporal/polyfill`. Neither specifier is in the hash `imports` map (and the playground’s own `v0CdnImports` also lacks `/date`). Selecting useDate and opening the playground fails module resolution even for the default `V0DateAdapter` path. Zip is fine — Vite honors package `exports`.

**Suggestion:** Emit every subpath `generateMainTs` can import (`@vuetify/v0/date`, `@vuetify/v0/notifications`, `@vuetify/v0/features/adapters/*`) plus `@js-temporal/polyfill` when Date is selected, using the same `dist/<subpath>/index.mjs` layout as `packages/0/package.json` `publishConfig.exports`. Drift-test generateImports against the `from '` specifiers in generated main.ts.

apps/builder/src/engine/manifest.ts
```diff
@@ -0,0 +1,1620 @@
...
+// `adapter` is listed wherever the form stores one: prune runs before the
+// adapter handler reads it, so omitting it here silently drops the adapter.
+export const KEYS: Record<string, string[]> = {
+  useBreakpoints: ['mobileBreakpoint', 'breakpoints'],
+  useDate: ['adapter', 'locale', 'locales', 'firstDayOfWeek'],
+  useFeatures: ['adapter', 'features'],
+  useHydration: [],
+  useLocale: ['default', 'fallback', 'messages', 'persist'],
+  useLogger: ['adapter', 'level', 'prefix', 'enabled'],
+  useNotifications: ['adapter', 'namespace', 'timeout'],
```
johnleider: **[suggestion]** v0 PluginOptions currently expose `persist?: boolean` on useTheme, useLocale, useRtl, useReducedMotion, useFeatures, and useNotifications. `persist: true` throws inside `createPluginContext` if `v0:storage` is not installed — which is why `selectedPluginIds` auto-inserts storage. The builder only offers persist on locale and reduced-motion. KEYS drops it for useFeatures (`['adapter', 'features']`), useNotifications (`['adapter', 'namespace', 'timeout']`), and useRtl (`['default', 'target']`). Theme KEYS lists `persist` but ThemeConfig never saves it, so generated apps still lose the selected theme on reload. The test “should allow persist for every plugin whose form offers it” only greps builder `*Config` interfaces, so it stays green while four of six persist-capable plugins are missing.

**Suggestion:** Add persist to the four forms and to KEYS (ThemeConfig snapshot must actually write it). Point the drift test at v0’s `*PluginOptions` (`persist?: boolean` on the interface), not the builder’s Config copies.

apps/builder/src/engine/manifest.test.ts
```diff
@@ -0,0 +1,788 @@
...
+  })
+
+  it('should import notification adapters from the notifications subpath', () => {
+    const source = generateMainTs(
+      ['useNotifications'],
+      { useNotifications: { adapter: 'KnockNotificationsAdapter' } },
+    )
+
+    expect(source).toContain(`import { KnockNotificationsAdapter } from '@vuetify/v0/notifications'`)
+  })
```
johnleider: **[suggestion]** `should import notification adapters from the notifications subpath` (and the matching features-adapter cases) assert `source.toContain(\`import { KnockNotificationsAdapter } from '@vuetify/v0/notifications'\`)`. BYOC adapters are not imported — `byoc()` returns `imports: []` and only writes that line inside a `// TODO:` comment. The test passes against the comment. A regression that emitted `new KnockNotificationsAdapter()` with no client, or that stopped importing from the real subpath in a non-comment position, would still pass.

**Suggestion:** Assert the TODO comment explicitly (`expect(source).toContain('//   import { KnockNotificationsAdapter }')`) and `expect(source).not.toMatch(/^import \{ KnockNotificationsAdapter \}/m)`. Drive the real import-path check off `BYOC[adapter].from`, not a substring of the whole file.

apps/builder/src/engine/manifest.ts
```diff
@@ -0,0 +1,1620 @@
...
+function reference (token: string): string {
+  const fallback = TOKENS[token]
+  return fallback ? `var(--v0-${token}, ${fallback})` : `var(--v0-${token})`
+}
+
+// Maps each theme token onto the CSS variable useTheme writes. Without this the
+// utility classes in the demos resolve to nothing and the app renders unstyled.
+export function generateUnoConfig (pluginConfig: Record<string, unknown> = {}): string {
+  const colors = collectTokens(pluginConfig)
+    .map(token => `      '${token}': '${reference(token)}',`)
```
johnleider: **[suggestion]** `generateUnoConfig` interpolates user theme token names and values into TS/CSS with ` '${token}': '${reference(token)}' `. ThemeConfig.vue accepts any non-empty color name. A token containing `'` or `\` breaks `uno.config.ts`; a token that is not a valid CSS custom-property ident (`foo bar`) emits `--v0-foo bar`, which is invalid CSS. Unlike `formatRestFields`, this path does not JSON-stringify.

**Suggestion:** Emit `JSON.stringify(token)` as the Uno key and sanitize the CSS variable name (reject or slug anything outside `[A-Za-z0-9_-]`). Hex values from the color picker are already safe; names are not.

apps/builder/src/engine/zip.ts
```diff
@@ -0,0 +1,213 @@
...
+  pluginConfig: Record<string, unknown>
+  selectedComponents: Set<string>
+  /** Composables pulled in by the resolver for the user's selection. */
+  resolved?: string[]
+}
+
+// Pinned as a mutually compatible set and verified against a real
+// `pnpm install && pnpm run build` of the generated starter.
+const DEPS: Record<string, string> = {
+  '@vuetify/v0': '^1.0.2',
```
johnleider: **[suggestion]** Generated `package.json` pins `'@vuetify/v0': '^1.0.2'` while this repo is 1.2.1 and codegen targets current APIs (`createDataTable` `onboard` / no `items` option, tooltip and reduced-motion plugins, etc.). `^1.0.2` installs 1.2.1 today, so a fresh `pnpm install` is fine, but the declared floor is a compatibility lie — a resolver that picked 1.0.2 would not compile these starters.

**Suggestion:** Pin `^1.2.1` (or whatever version the “install && build” verification actually used). Re-verify when cutting a builder release.

PR #269, opened 2026-06-03
chore: enforce V0Error over bare Error in packages/0/src via no-restricted-syntax

### LINKED ISSUE #248: [Feature Request] Enforce V0Error on throws in packages/0/src via eslint-plugin-vuetify
## Context

PR #247 adopts a typed \`V0Error\` class for every throw in \`packages/0/src\` so consumers can identify v0 errors via a stable \`code\` discriminant. Coverage is currently 8/8 throw sites, but the convention is enforced by author discipline only — nothing stops a future PR from landing a bare \`throw new Error(...)\`.

## Ask

Add a lint rule to \`eslint-plugin-vuetify\` (or a project-local rule under \`packages/0/.eslintrc\`) that flags bare \`throw new Error(...)\` inside \`packages/0/src/**/*.ts\` and suggests \`throw new V0Error(...)\` with a code from the \`V0ErrorDetails\` union.

## Suggested semantics

- **Target:** \`packages/0/src/**/*.ts\` excluding \`*.test.ts\` and \`*.bench.ts\`.
- **Detect:** \`ThrowStatement > NewExpression[callee.name='Error']\`.
- **Carve-outs:**
  - Test/bench files (already excluded by path).
  - Re-throws (\`throw err\` where \`err\` is a caught binding).
  - Throws inside test-helper utilities used by tests (rare).
- **Autofix:** none — the author has to pick a code from \`V0ErrorDetails\`, and adding a new arm to the union is a deliberate decision.
- **Message:** \`Use V0Error with a code from V0ErrorDetails. See https://github.com/vuetifyjs/0/blob/master/packages/0/src/utilities/errors.ts.\`

## Notes

- This is the enforcement gap flagged by the type-design review on #247 ("structural cause shape *cannot* be safely tested without inspecting \`code\` — that's the brand").
- Without the rule, the registry decays — future authors won't know it exists or will skip it for a bare \`Error\` when they're in a hurry.

sridhar-3009:
## Problem

Closes #248.

PR #247 converted all 8 existing `throw new Error(...)` sites in `packages/0/src` to use `V0Error` with a typed code from `V0ErrorDetails`. The convention was described in the philosophy but enforced only by author discipline — nothing in CI would catch a future PR landing a bare `throw new Error(...)`.

## Fix

Add a `no-restricted-syntax` block in `eslint.config.js` scoped to `packages/0/src/**/*.ts` (with carve-outs for `*.test.ts` and `*.bench.ts`):

```js
{
  name: 'vuetify/no-bare-error-throw',
  files: ['packages/0/src/**/*.ts'],
  ignores: ['packages/0/src/**/*.test.ts', 'packages/0/src/**/*.bench.ts'],
  rules: {
    'no-restricted-syntax': ['error',
      {
        selector: 'ThrowStatement > NewExpression[callee.name="Error"]',
        message: 'Use V0Error with a code from V0ErrorDetails instead of bare Error. See packages/0/src/utilities/errors.ts.',
      },
    ],
  },
},
```

## Notes

- Uses `no-restricted-syntax` — no custom plugin, no new dependency, mirrors the existing `withDefaults` guard on line 29.
- Re-throws (`throw err` where `err` is a caught binding) are not flagged — the selector only matches `new Error(...)` constructor calls.
- Test and bench files are excluded per the issue spec.
- The rule fires at `error` severity so it blocks CI on a violation.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -127,6 +127,22 @@ export default vuetify({
   ignores: ['**/export-templates/package.json'],
 },
 {
+  // Enforce V0Error over bare Error for throw sites in the core package.
+  // This prevents contributors from accidentally bypassing the typed error
+  // registry (V0ErrorDetails) that gives consumers a stable code discriminant.
+  name: 'vuetify/no-bare-error-throw',
+  files: ['packages/0/src/**/*.ts'],
+  ignores: ['packages/0/src/**/*.test.ts', 'packages/0/src/**/*.bench.ts'],
+  rules: {
+    'no-restricted-syntax': ['error',
+      {
+        selector: 'ThrowStatement > NewExpression[callee.name="Error"]',
+        message: 'Use V0Error with a code from V0ErrorDetails instead of bare Error. See packages/0/src/utilities/errors.ts.',
+      },
+    ],
+  },
+},
+{
   name: 'vuetify/no-v0-imports',
   files: ['packages/paper/**/*.{ts,js,vue}', 'app/**/*.{ts,js,vue}'],
   rules: {


johnleider:
Thank you for your contribution and interest in improving Vuetify0!

PR #276, opened 2026-06-06
fix(locale): English fallback for unconfigured consumers (WCAG 4.1.2)

### LINKED ISSUE #196: locale.t() needs a key-level fallback — WCAG 4.1.2 gap
## Problem

After [9823d09b](../commit/9823d09b) (triage row B07), six components route aria strings through `locale.t()`:

- `Dialog/DialogClose.vue` — `locale.t('Dialog.close')`
- `AlertDialog/AlertDialogClose.vue` — `locale.t('AlertDialog.close')`
- `Snackbar/SnackbarClose.vue` — `locale.t('Snackbar.close')`
- `NumberField/NumberFieldIncrement.vue` — `locale.t('NumberField.increment')`
- `NumberField/NumberFieldDecrement.vue` — `locale.t('NumberField.decrement')`
- `Rating/RatingRoot.vue` — `locale.t('Rating.valueText', { value, size })`

If a consumer hasn't registered locale messages (and v0 ships none by design), `locale.t('Dialog.close')` returns the literal string `"Dialog.close"` as the aria-label. That fails WCAG 4.1.2 Name, Role, Value — assistive tech announces the key identifier instead of a human label.

The same hole applies to `createLocaleFallback()` (the no-adapter path), which also returns the key verbatim.

## History

Commit [245182f0](../commit/245182f0) (2026-03-20, \`feat(useLocale): change t() to variadic signature\`) removed the \`fallback?\` parameter from \`t()\` to align with Vuetify's \`t(key, ...params)\` convention. The BREAKING CHANGE note advised consumers to use the fallback locale mechanism or ship default messages — but v0 itself ships no defaults and components have no way to contribute their own.

## Rejected approaches (2026-04-20)

- **Re-add the fallback param** (\`t(key, params?, fallback?)\`) — reverses the 2026-03-20 BREAKING CHANGE, re-introduces divergence from Vuetify's signature.
- **v0 ships minimal default English messages at plugin install** — violates headless minimalism; v0 would own strings it currently doesn't.

## Angles worth exploring

- Component-level \`ariaLabel\` prop as the fallback boundary, so consumers supply English once without wiring up \`useLocale\`. Fallback chain becomes \`prop ?? locale.t(key) ?? ???\`.
- Missing-key handler on the adapter — \`createLocale({ onMissing: (key) => ... })\`. Pushes the decision to app layer.
- A \`@vuetify/v0/messages/en\` subpath export that consumers can \`app.use(createLocalePlugin({ messages: { en } }))\` — headless by default, but defaults are reachable without re-implementing them.

## Tension

Three competing pulls, all load-bearing:

1. Vuetify signature alignment (for direct adapter migration)
2. Ergonomic safety for unconfigured consumers (WCAG 4.1.2)
3. Headless minimalism (v0 ships no strings)

Today (1) and (3) are satisfied; (2) is the gap.

## Acceptance

An approach that closes the WCAG gap without reverting (1) or violating (3). The two rejected options above are off the table.

sridhar-3009:
## Problem

When `useLocale()` is called without a locale plugin (e.g. `createLocalePlugin()` not installed), `createLocaleFallback` returned the raw dot-notation key as the string — e.g. `t('Dialog.close')` → `'Dialog.close'`. This violates WCAG 4.1.2 (Name, Role, Value): screen readers announce `"Dialog.close"` instead of a human-readable label.

Closes #196

## Solution

Add an `@vuetify/v0/locale/messages/en` subpath export containing all English component strings. `createLocaleFallback` now looks up keys against that dictionary and returns the English translation, so `t('Dialog.close')` → `'Close'` even with no locale plugin.

Also adds an `onMissing` hook to `LocaleOptions`, `LocaleAdapterContext`, and `V0LocaleAdapter`, giving configured consumers a way to supply custom fallbacks when a key is missing from their locale.

## Changes

- `packages/0/src/locale/messages/en/index.ts` — new: English strings for all component aria keys
- `packages/0/package.json` — add `./locale/messages/en` conditional export
- `packages/0/src/composables/useLocale/index.ts` — `createLocaleFallback` uses `lookupEn()` + `_interpolate()`; add `onMissing` to `LocaleOptions`
- `packages/0/src/composables/useLocale/adapters/adapter.ts` — add `onMissing?` to `LocaleAdapterContext`
- `packages/0/src/composables/useLocale/adapters/v0.ts` — call `onMissing` before raw-key fallback
- `packages/0/src/composables/useLocale/index.test.ts` — 9 new tests: `onMissing` hook, `createLocaleFallback` English defaults
- Updated component tests (Avatar, Breadcrumbs, NumberField, Pagination) to expect English strings instead of raw keys

## Test plan

- [ ] `pnpm test:run` — 151 files, 6110 tests pass
- [ ] `pnpm typecheck` — no errors
- [ ] `pnpm repo:knip` — no unused exports
- [ ] Without locale plugin: `t('Dialog.close')` → `'Close'`; `t('Pagination.status', { page: 2, pages: 10 })` → `'Page 2 of 10'`
- [ ] With locale plugin: existing behavior unchanged; `onMissing` fires only for missing keys after fallback chain


johnleider (commented):
Maintainer pass on the locale code. (Context: this exact commit accidentally rode into #277 off a shared base — that PR's been stripped back to just its playground fix, so this is the feature's proper home.) Four findings — I'm pushing a follow-up commit to this branch that fixes 1–3; **#4 needs your decision**.

**1 — Duplicated interpolation + inaccurate comment** (`useLocale/index.ts`)
`_interpolate` is a near-verbatim copy of `V0LocaleAdapter.interpolate`, and the comment *"Reuse V0LocaleAdapter interpolation by creating a minimal adapter instance"* describes neither what the code does (no instance is created) nor reuse (the logic is re-implemented). Root cause: the adapter's `interpolate` is `private`, so the module-level `lookupEn` can't reach it. Fix: extract one shared `interpolate(message, args)` helper that both the adapter and `lookupEn` call — removes the duplication and the drift risk (two copies of the placeholder grammar will diverge).

**2 — Raw type checks instead of `#v0/utilities` guards** (`useLocale/index.ts`)
`lookupEn`/`_interpolate` use `node === null || typeof node !== 'object'`, `typeof node !== 'string'`, `typeof rest[0] === 'object'`. The repo standard — and the adapter one file over — uses `isNull`/`isObject`/`isString`/`isUndefined`. Switched to those.

**3 — Broken `onMissing` JSDoc example** (`useLocale/index.ts`)
```ts
createLocalePlugin({ onMissing: key => en[key] })
```
This can't work: `en` is nested (`en.Dialog.close`), so `en['Dialog.close']` is `undefined` for every dotted key. Replaced with a correct flat-dictionary example that matches the `(key) => string | undefined` contract.

**4 — Tests now assert exact English strings — your call** (`Avatar`/`Breadcrumbs`/`NumberField`/`Pagination` specs)
The component tests were flipped from raw-key/`toBeDefined` to exact English (`'+3 more'`, `'Increment'`, `'Page 2 of 10'`, `'Breadcrumbs'`). PHILOSOPHY line 588 says *"Tests assert `toBeDefined()`, not exact strings"* precisely so component wording can change without breaking tests. Counterpoint: these tests now exercise the en-bundle feature specifically, where asserting the shipped string is arguably legitimate. I left them untouched — decide whether to keep the exact assertions (and add a line-588 exception note) or loosen to `expect(label).not.toBe(key)`.

Net: the behavior change itself (`createLocaleFallback` raw-key → English) is a reasonable WCAG 4.1.2 win; it's a documented-behavior change, so it's worth a PHILOSOPHY §5.5 note in the same PR.

johnleider:
Superseded by #372, which fixed the underlying WCAG 4.1.2 gap (#196) via the `ti()` + inline-component-defaults approach instead of this bundled-English-dictionary design. The `ti()` rewrite (commit 2b290474, "replace bundled-English fallback with ti() + inline component defaults") deliberately replaced this PR's mechanism. Closing as superseded.

PR #283, opened 2026-06-09
docs(guide): add dedicated testing guide page

### LINKED ISSUE #249: [Feature Request] Dedicated testing guide page
## Context

PR #247 introduces \`V0Error\` / \`isV0Error\` as the canonical way to assert v0-thrown errors in user tests. There's no docs page today that covers testing patterns for v0 consumers — testing knowledge is scattered across \`.claude/rules/testing.md\` (which is contributor-private) and individual test files.

## Ask

Add a dedicated testing guide page to \`apps/docs\`, probably under a new "Guides" section or a "Recipes / Testing" subsection.

## Suggested scope

- **Setup** — recommended deps (Vitest, happy-dom or jsdom), config snippets.
- **Mocking Vue DI** — \`vi.mock('vue', () => ({ provide, inject }))\` pattern + \`hasInjectionContext\` extension.
- **Mocking plugins** — \`global.plugins: [createThemePlugin(), ...]\` in @vue/test-utils.
- **Fake timers** — \`vi.useFakeTimers()\` + \`vi.advanceTimersByTime\` for \`useTimer\`, \`useDelay\`, \`usePopover\`, etc.
- **SSR-only tests** — splitting \`*.ssr.test.ts\` with hoisted \`vi.mock('#v0/constants/globals', ...)\`.
- **Asserting throws + cause shape** — \`isV0Error(err, 'V0_CONTEXT_MISSING')\` narrowing pattern, full list of codes from \`V0ErrorDetails\`.
- **Asserting warnings** — \`vi.spyOn(console, 'warn')\` capture-and-assert pattern (PHILOSOPHY §9.2 says never silently swallow).
- **Locale-safe assertions** — \`toBeDefined()\` not exact text (PHILOSOPHY §5.5).

## Reference material

- \`packages/0/.claude/rules/testing.md\` — current contributor-facing rules; a lot of this content can be lifted to a user-facing page.
- Example tests across \`packages/0/src/composables/**/*.test.ts\`.

## Notes

- Originated as side feedback during PR #247 review — the dedicated page is meant to be the *one place* users go when they want to test v0-based code, including the new \`isV0Error\` pattern.

sridhar-3009:
Closes #249

Adds `/guide/tooling/testing` — a consumer-facing testing guide covering:

- **Setup** — Vitest + happy-dom, `vitest.config.ts` snippet
- **Mocking Vue DI** — `vi.mock('vue', ...)` pattern + `hasInjectionContext` extension
- **Mocking plugins** — `global.plugins: [createStackPlugin(), createThemePlugin()]`
- **Fake timers** — `vi.useFakeTimers()` + `vi.advanceTimersByTime` for timer/delay/popover composables
- **Asserting v0 errors** — `isV0Error(err, code)` narrowing with a table of all `V0ErrorCode` variants
- **Asserting warnings** — `vi.spyOn(console, 'warn')` with `using` auto-restore
- **SSR-only tests** — `*.ssr.test.ts` split + hoisted `vi.mock('#v0/constants/globals')`
- **Locale-safe assertions** — `toBeDefined()` over exact text values

Content is adapted from the contributor-private `.claude/rules/testing.md` into a user-facing reference.

### DIFF
--- a/apps/docs/src/pages/guide/tooling/testing.md
+++ b/apps/docs/src/pages/guide/tooling/testing.md
@@ -39,6 +39,10 @@ npm install -D vitest @vue/test-utils happy-dom
 yarn add -D vitest @vue/test-utils happy-dom
 ```
 
+```bash bun
+bun add -D vitest @vue/test-utils happy-dom
+```
+
 :::
 
 Configure Vitest to use happy-dom:
--- a/apps/docs/src/pages/guide/tooling/testing.md
+++ b/apps/docs/src/pages/guide/tooling/testing.md
@@ -12,7 +12,7 @@ meta:
 related:
   - /guide/tooling/ai-tools
   - /guide/tooling/vuetify-cli
-  - /guide/essentials/types
+  - /guide/features/types
 ---
 
 # Testing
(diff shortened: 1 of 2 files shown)


johnleider:
Thanks for this — solid guide. I pushed two fixes directly to the branch:

**1. The "Asserting v0 errors" example didn't actually throw.** It called `useTheme()` expecting a `V0Error`, but `useTheme` ships a fallback (`createPluginContext(..., { fallback: () => createThemeFallback() })`) and returns synthesized defaults when the plugin is absent instead of throwing — so `caught` stays `undefined` and the test fails at `expect(caught).toBeInstanceOf(V0Error)`. Switched the example to `useDate()`, which genuinely throws `V0_PLUGIN_MISSING` when no provider is installed, and narrowed to its real `.plugin` payload (the `.key` field only exists on `V0_CONTEXT_MISSING`). Added a short note that fallback-backed composables like `useTheme` don't throw.

**2. Broken `related` link.** `/guide/essentials/types` doesn't exist — the Types page lives at `/guide/features/types`. Fixed.

The rest checked out against source: the error-code table is exact, `isV0Error(err, code)` narrowing is correct, and `V0Error`/`isV0Error`/`createStackPlugin`/`createThemePlugin`/`useTimer` are all genuinely exported from `@vuetify/v0`.

Two non-blocking notes, no change needed:
- The SSR example's `.toThrow('[v0] useDate() must be called inside a Vue component')` is a truncated prefix of the real message (`…with createDatePlugin installed.` plus an example block), but `.toThrow(string)` is substring-matching so it still passes.
- In that same SSR example the `IN_BROWSER: false` mock isn't what triggers the throw — `useDate` gates on `instanceExists()` (being called outside a component), not on `IN_BROWSER`.

johnleider (approved):

PR #287, opened 2026-06-09
perf(createDataTable): skip per-row shallowReactive proxy via reactiveTickets option

### LINKED ISSUE #257: [Bug Report] createDataTable: creation and pipeline performance regressed after #236
## Summary

A benchmark regen after #236 shows a broad performance regression in `createDataTable` construction and the read pipeline. The stable, low-variance benchmarks:

| Benchmark | Before | After | Latency |
|---|---|---|---|
| Create table (10,000 items) | blazing | **slow** | ~25 ms |
| Create table with all options (1,000 items) | fast | **slow** | ~117 ms |
| Create with openAll (1,000 items) | fast | **slow** | ~107 ms |
| Compute groups (10,000 items) | fast | **slow** | ~33 ms |
| `initialization` group | blazing | good | — |

The 10k sort/search pipeline benches also dropped to `slow`, but they have high run-to-run variance — treat their absolute numbers cautiously. The creation benches above are stable (<2% rme) and are the reliable signal.

## Root cause

#236 dropped the bulk `items` option in favour of registry `onboard`. Two consequences at construction time:

- The **row** registry is created with `reactive: true` (`packages/0/src/composables/createDataTable/index.ts:375`), so every row ticket is wrapped in a `shallowReactive` proxy during `register`/`onboard` — 10k proxy allocations for a 10k-row table.
- The adapter `source` is now `computed(() => registry.values().map(t => t.value))` (`index.ts:546`), rebuilding a fresh array via an uncached `Array.from` on every pipeline read.

The old path passed `items` straight to the adapter by reference, with zero per-row work at creation.

## Suggested direction

- Drop `reactive: true` on the **row** registry (keep it on columns). The pipeline reads rows through the `source` computed, so per-row template reactivity buys nothing the table actually consumes.
- Failing that, give `createRegistry.onboard` a batched fast path that wraps the collection reactively once instead of per-ticket, and memoize `registry.values()` so `source` isn't an uncached `Array.from` per read.

Bench data regenerated on `master`; see `apps/docs/src/data/metrics.json`.

sridhar-3009:
Closes #257

## Problem

After #236 replaced the bulk `items` option with registry `onboard`, a 10 k-row
table allocates 10 k `shallowReactive` proxies at construction — one per ticket.
The proxy allocation is the dominant cost in the regression benchmarks:

| Benchmark | Before | After this PR |
|---|---|---|
| Create table (10,000 items) | slow (~25 ms) | eliminates per-row proxy cost |
| Create with all options (1,000 items) | slow (~117 ms) | eliminates per-row proxy cost |
| Create with openAll (1,000 items) | slow (~107 ms) | eliminates per-row proxy cost |

## Solution

Adds `reactiveTickets?: boolean` to `RegistryOptions`. When `false`, the **collection**
(shallowReactive Map + shallowReactive order array) stays reactive so that computed
derivations like `source` re-evaluate when rows are added or removed, but individual
tickets are plain objects instead of Proxy wrappers.

```ts
// createRegistry options type — new field:
reactiveTickets?: boolean  // defaults to same as `reactive`
```

Sets `reactiveTickets: false` on the **row** registry in `createDataTable`.
Column tickets remain fully reactive because columns can be mutated in place
(sort state, width, visibility).

The key invariant: `source = computed(() => registry.values().map(t => t.value))`
reads from `order` which is still `shallowReactive`, so the computed re-triggers
on any register/unregister. Per-ticket reactivity was never consumed by the pipeline.

## Changes

- `packages/0/src/composables/createRegistry/index.ts` — add `reactiveTickets` option, default to `reactive`; use it in the `register()` ticket-wrapping path
- `packages/0/src/composables/createDataTable/index.ts` — pass `reactiveTickets: false` for the row registry

All 6108 tests pass.

### DIFF
--- a/packages/0/src/composables/createDataTable/index.ts
+++ b/packages/0/src/composables/createDataTable/index.ts
@@ -375,6 +375,7 @@ export function createDataTable<T extends Record<string, unknown>> (
   const registry = createRegistry<DataTableTicketInput<T>, DataTableTicket<T>>({
     events: true,
     reactive: true,
+    reactiveTickets: false,
   })
 
   const columns = createRegistry<DataTableColumnTicketInput<T>, DataTableColumnTicket<T>>({
--- a/packages/0/src/composables/createRegistry/index.ts
+++ b/packages/0/src/composables/createRegistry/index.ts
@@ -726,6 +726,17 @@ export interface RegistryOptions {
    * ```
   */
   reactive?: boolean
+  /**
+   * Wrap each registered ticket in `shallowReactive` so that mutations to its
+   * properties are observable by Vue effects.
+   *
+   * @default same as `reactive`
+   * @remarks Set to `false` together with `reactive: true` when you need the
+   * collection to be observable (adds/removes re-trigger computed) but do not
+   * need per-ticket property reactivity. This avoids the proxy-allocation cost
+   * for large sets of items that are read, not mutated in place.
+   */
+  reactiveTickets?: boolean
 }
 
 export interface RegistryContextOptions extends RegistryOptions {
@@ -763,6 +774,7 @@ export function createRegistry<
 
   const events = options?.events ?? false
   const reactive = options?.reactive ?? false
+  const reactiveTickets = options?.reactiveTickets ?? reactive
 
   const collection = reactive ? shallowReactive(new Map<ID, E>()) : new Map<ID, E>()
   const catalog = new Map<unknown, ID[]>()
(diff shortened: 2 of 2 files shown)


johnleider:
Closing in favor of a re-cut of #274's approach. Benchmarked head-to-head against the #257 repro: dropping only the per-ticket proxies recovers ~35% (10k-row creation 43.3ms → 28.9ms — still in the slow tier the issue reports) because the dominant cost is the shallowReactive collection, not the ticket proxies. The replacement also avoids the `reactiveTickets` situation where a registry advertises `reactive: true` while upsert propagation (PHILOSOPHY §4.4) is silently voided — `upsert(id, { value })` stops re-running the pipeline under this branch. Replacement PR incoming with the regression coverage both branches were missing.

PR #289, opened 2026-06-09
feat(useReducedMotion): add reduced-motion plugin with system/always/never mode override

### LINKED ISSUE #252: [Feature Request] Reduced-motion override
related to [vuetify#22159](https://github.com/vuetifyjs/vuetify/issues/22159)

The request: add global config so users (and apps) can override `prefers-reduced-motion`, exposed via a hook (`useReducedMotion()`) and a data attribute on `<body>` for CSS.

### Does it fit v0?

v0 already ships `usePrefersReducedMotion()` utility and the plugin would be a natural extension.

### What needs to be covered

- new plugin - similar to useTheme (in how it resolves `prefers-color-scheme`)
- transition tokens that read the resolved value (expected in Paper or Vuetify)

Logic input value: `mode: 'system' | 'always' | 'never'` (reactive).
- `'system'` means resolving media query - same challenge with SSR as for adapting to "system" theme (requires [client-hints](https://wicg.github.io/user-preference-media-features-headers/))
- `'always'/'never'` - user/app preference

Paper/Vuetify are meant to consume the `reduced` boolean flag
- JS-based animations could pick the readonly `ref` from composable or `provide/inject`
- CSS-based transitions could be branch out behind `[data-reduced-motion="reduce"]` to either apply fade in/out or skip transition entirely

### Challenges / limitations

- **CSS coordination** v0 is headless, so the *behavior* is just a flag. Anything that actually drops a transition lives in Paper / Vuetify / app CSS. The plugin only sets the attribute; consumers must author `@media (prefers-reduced-motion: reduce)` *and*  selectors.
- **Granularity** ON/OFF `[data-reduced-motion="reduce"]` and `const { current } = useReducedMotion()` are meant to be consumed/translated into "keep fades, drop transforms" semantics. VPaper or SCSS/CSS tokens could define the policy.
- **SSR + hydration** first paint may still flicker unless the app injects tokens server-side
- **Naming**: `system/always/never` (clearer than in Motion `user/always/never`)

### Recommendation

Land `useReducedMotion` in v0 as a `createPlugin`-style provider (mirroring `useTheme`), exposing `{ mode, reduced, current }` and side-effecting `document.body.dataset.reducedMotion`. Keep policy (what to fade, what to drop) in Paper/Vuetify. The Vuetify-side regression that prompted the issue (v3.10.0) is a separate fix that consumes this primitive.

---

Sources:

- [Motion — Config](https://motion.dev/docs/react-motion-config#reducedmotion)
- [Motion — useReducedMotion](https://motion.dev/docs/react-use-reduced-motion)
- [web.dev — prefers-reduced-motion](https://web.dev/articles/prefers-reduced-motion)

sridhar-3009:
Closes #252

## Summary

Adds `useReducedMotion` — a plugin-style composable that bridges the system
`prefers-reduced-motion` media query with an application-level override, following
the same `createPluginContext` trinity pattern used by `useTheme`, `useRtl`, and
`useLocale`.

## API

```ts
import { createReducedMotionPlugin, useReducedMotion } from '@vuetify/v0'

app.use(createReducedMotionPlugin({ default: 'system' }))

// inside a component or composable
const { mode, reduced, current } = useReducedMotion()
mode.value = 'always'          // force-reduce regardless of system setting
console.log(reduced.value)     // true
console.log(current.value)     // 'reduce'
```

| Property | Type | Description |
| --- | --- | --- |
| `mode` | `Ref<'system' \| 'always' \| 'never'>` | Writable — controls the override |
| `reduced` | `Readonly<ComputedRef<boolean>>` | True when motion should be reduced |
| `current` | `Readonly<ComputedRef<'no-preference' \| 'reduce'>>` | CSS-keyword form |

## Behaviour

- `'system'` — mirrors the `prefers-reduced-motion: reduce` media query (read synchronously on mount, live-updated via a `change` listener in the adapter)
- `'always'` — forces `reduced = true` regardless of system setting
- `'never'` — forces `reduced = false` regardless of system setting
- `V0ReducedMotionAdapter` sets `document.body.dataset.reducedMotion` to `'reduce'` or `'no-preference'` so that CSS can branch on `[data-reduced-motion="reduce"]` without depending on a media query

## Files

- `packages/0/src/composables/useReducedMotion/index.ts` — composable + plugin trinity
- `packages/0/src/composables/useReducedMotion/adapters/adapter.ts` — abstract adapter interface
- `packages/0/src/composables/useReducedMotion/adapters/v0.ts` — default adapter (body dataset + media query listener)
- `packages/0/src/composables/useReducedMotion/index.test.ts` — 9 tests covering mode switching, adapter side-effects, and media query reactivity
- `packages/0/src/composables/index.ts` — export

All 6119 tests pass.


johnleider:
closed by #278

PR #337, opened 2026-06-12
feat(composables): add useReducedMotion plugin

johnleider:
## Description

Adds the `useReducedMotion` plugin composable: respect or override the user's `prefers-reduced-motion` preference with a reactive `isReduced` flag, a three-way `mode` (`system` / `always` / `never`), optional persistence, and a default adapter that writes `data-reduced-motion` on `document.body` in the browser and renders it via `@unhead` `bodyAttrs` under SSR — reactively, via `entry.patch`.

Closes #278

## Highlights

- **Leak-safe lifecycle** — the context exposes `dispose()`; the plugin registers it and the adapter's late-bound `dispose` on `app.onUnmount`, so teardown works even though plugin install runs outside any effect scope.
- **Honest SSR story** — the server renders the *configured* mode (the OS preference is only readable client-side); the docs recommend pairing the attribute selector with a native `@media (prefers-reduced-motion: reduce)` rule for pre-hydration coverage.
- **Typed unhead seam** — structural `Head`/`HeadEntry` duck-types; no `@unhead` dependency in package source.
- **Persistence** — `persist: true` stores the mode through the storage plugin, with a validated `restore` (no blind casts).
- Docs page with architecture diagram and robust reduce CSS, accessible example (`Button.Group` mode switcher), full test coverage including SSR persist/restore ordering, and registry entries (`maturity.json` preview, READMEs, barrels).
- `.claude/rules/` adapter-pattern snippets were found stale against all 15 real adapter directories during review — corrected in this PR, with plugin conventions (options tiers, restore validation, unhead resolution, fallback naming) codified.


johnleider:
Duplicate of #278 — opened against the wrong branch name; the four new commits are now pushed to feat/reduced-motion-plugin-252 and live in #278.

PR #343, opened 2026-06-12
fix(createPlugin): defer factory call to install time to prevent resource leaks

### LINKED ISSUE #338: fix(createPlugin): plugin contexts are created eagerly at factory time, stranding live resources
## Problem

`createXPlugin()` (via `createPluginContext`) runs the context factory at **plugin-creation time**, not at install time:

```ts
// createPlugin/index.ts — createXPlugin
const [, provide, context] = createXContext({ ...options, namespace })  // factory runs HERE
```

Factories that allocate live resources attach them before `app.use()` ever runs. Concrete case: `createReducedMotionPlugin()` in the browser attaches a `matchMedia` listener at the factory call — `useMediaQuery`'s `immediate` watch fires because `useHydration()` resolves its fallback (`isHydrated: true`) outside any injection context.

## Consequences

- A plugin **created but never installed** leaks its listener permanently — nothing ever runs `setup`, so no `app.onUnmount` teardown is registered.
- Two plugin instances with the **same namespace** on one app: the second install is silently skipped by the `INSTALLED` guard, its `setup` never runs, and its context's resources are stranded with no reachable handle.
- `useMediaQuery`'s "defer the change listener until after hydration" contract silently never engages on the plugin path — the context is created outside injection context, so it always sees the hydration fallback.

This is structural to `createPluginContext` — every plugin whose factory allocates listeners/observers inherits it.

## Proposed direction

Lazy context creation: run the factory inside `provide` on first install. Needs care for the public `createXContext` path and the persist/restore ordering (`restore` runs in `provide` before `setup` — that ordering must survive).

Found during the #278 review.

sridhar-3009:
Fixes #338

## Problem

`createXPlugin()` (via `createPluginContext`) called the context factory **at plugin-creation time** — before `app.use()` ever runs:

```ts
// Before: factory runs here, at createXPlugin() call time
const [, provide, context] = createXContext({ ...options, namespace } as O)
```

This means:
- A plugin **created but never installed** allocates live resources (matchMedia listeners, watchers, unhead entries) that are never torn down
- Two plugin instances with the **same namespace** on one app: the second install is silently skipped, but its factory already ran and its resources are stranded
- `useMediaQuery`'s hydration-deferral contract silently breaks on the plugin path — the factory runs outside any injection context

## Solution

Move factory invocation inside the `provide` callback so it runs at install time, not factory time:

```ts
// After: factory deferred to install time
let context!: E

return createPlugin({
  namespace,
  provide: app => {
    const [, provide, ctx] = createXContext({ ...options, namespace } as O)
    context = ctx
    provide(context, app)
    // ...restore...
  },
  setup: app => {
    // context is set by provide, which always runs first
    config?.setup?.(context, app, options)
    // ...persist watch...
  },
})
```

`provide` always runs before `setup` (enforced by `createPlugin`), so the definite assignment is safe. The `INSTALLED` guard ensures `setup` never runs without a prior `provide` call.

`createXContext` (the standalone path) is unchanged — standalone contexts still create eagerly, which is correct.

## Test plan

- [ ] 2 new tests verify factory is not called at `createXPlugin()` time
- [ ] All existing persist/restore ordering tests pass
- [ ] 6110 tests pass, typecheck and lint clean

### DIFF
--- a/packages/0/src/composables/createPlugin/index.test.ts
+++ b/packages/0/src/composables/createPlugin/index.test.ts
@@ -240,6 +240,20 @@ describe('createPluginContext', () => {
     expect(factory).not.toHaveBeenCalled()
   })
 
+  it('should not run a second factory when a duplicate namespace is installed', () => {
+    const first = vi.fn(() => ({ value: 'first' }))
+    const second = vi.fn(() => ({ value: 'second' }))
+
+    const [, createFirst] = createPluginContext('v0:duplicate-namespace-test', first)
+    const [, createSecond] = createPluginContext('v0:duplicate-namespace-test', second)
+
+    createFirst().install(mockApp)
+    createSecond().install(mockApp)
+
+    expect(first).toHaveBeenCalledTimes(1)
+    expect(second).not.toHaveBeenCalled()
+  })
+
   it('should call setup callback on plugin install', () => {
     const setup = vi.fn()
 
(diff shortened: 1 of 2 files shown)


johnleider:
Pushed a small follow-up directly to this branch as a maintainer edit (d7e6f3cf), on top of your fix.

### What it addresses

The lazy-factory move is correct. One edge it left open is in `setup`, which still read the shared outer `context` binding. Now that the factory runs once **per install**, the same plugin *object* installed on more than one app reassigns `context` on each install — and the persist `watch` source closes over that mutable binding lazily, so a later install pulls the source out from under an earlier app's watcher.

Concretely: with a single `createXPlugin({ persist: true })` object used on two apps, mutating app A's state ends up persisting app B's value into app A's storage key.

Fix is one capture — `const ctx = context` at the top of `setup`, used for both the `config.setup` call and the persist `watch` — so each app's watcher tracks the instance that `provide` created for it. The adapter path was already safe (it receives `context` as a parameter, a fresh per-call binding); only the in-`createPluginContext` persist watch needed it.

### Tests added

- `should not run a second factory when a duplicate namespace is installed` — covers the issue's second scenario (a second same-namespace plugin object on one app must not run its factory).
- `should persist each app independently when one plugin object is reused` — fails against the shared-closure version (`'b-initial'` leaks into app A's storage) and passes after the capture.

### Also

Trimmed the two inline comment blocks down to the why.

Verified in `packages/0`: `vitest` 23/23 on `createPlugin`, `vue-tsc --noEmit` clean, ESLint clean on both files.

johnleider (approved):

PR #344, opened 2026-06-12
fix(NumberField): pass parsed value to commit() to avoid stale model read

### LINKED ISSUE #329: NumberField: typed value never commits via parent v-model — commit() clobbers the parse with a stale model read
### Environment
`packages/0` master (worktree at feat/builder, identical NumberField source).

### Repro
1. Bind a NumberField to parent state: `<NumberField.Root v-model="state.breakpoints[name]">` (e.g. the builder app's Breakpoints form, `apps/builder/src/plugins/breakpoints/BreakpointsConfig.vue`).
2. Focus the control, type a new value (e.g. `1000`), then blur or press Enter.
3. The field snaps back to the previous value; the parent model never receives the typed value.

ArrowUp/ArrowDown stepping and programmatic writes propagate fine — only the typed parse→commit path is broken, and only when the model is parent-bound (plain local refs take `defineModel`'s local-value path and don't reproduce, which is why unit tests pass).

### Mechanism
1. `NumberFieldControl.vue` `onInput` updates only `text.value`, not the model.
2. On blur/Enter (`NumberFieldControl.vue:102-108`, `:128-132`): `root.value.value = parsed` then `root.commit()`.
3. `root.value` is the `defineModel` ref (`NumberFieldRoot.vue:227`). With a parent listener, the write emits `update:modelValue`, but the child's prop hasn't round-tripped yet — a synchronous re-read returns the **stale** value.
4. `commit()` (`createNumberField/index.ts:208-219`) re-reads `value.value` synchronously and re-assigns `value.value = numeric.snap(value.value)` — re-emitting the **old** number and clobbering the parse in the same tick. `syncText()` then restores the old text.

The existing tests document the flake instead of catching it: `packages/0/src/components/NumberField/index.test.ts:1568` and `:1588` accept **either** the new or the old value.

### Proposed fix
Thread the parsed value through instead of re-reading the model:

```ts
// NumberFieldControl.vue (blur + Enter)
const parsed = root.parse(text.value)
root.value.value = parsed
root.commit(parsed)

// createNumberField/index.ts
function commit (next?: number | null): void {
  const val = next ?? value.value
  if (isNull(val)) return
  // ...range/snap logic on `val`...
  value.value = numeric.snap(val)
}
```

Regression test: mount with a parent-bound v-model, type + blur, assert the parent ref received the typed value (and tighten the two either/or assertions at `index.test.ts:1568`/`1588`).

Worth grepping sibling composables (createSlider, createInput, createRating) for the same write-then-synchronously-re-read-model shape while fixing.

### Impact
Any consumer binding NumberField through parent state cannot type values — discovered because custom breakpoint px values are impossible to enter in the framework builder.

sridhar-3009:
## Problem

When `onBlur` or the `Enter` key handler in `NumberFieldControl` assigns `root.value.value = parsed` and then calls `root.commit()` on the same tick, the parent-bound `v-model` hasn't round-tripped yet. `commit()` re-reads `value.value` and sees the stale pre-write value, so the snap/clamp result is discarded.

Repro: bind a `v-model` to a `NumberField`, type a new number, blur — parent model stays at old value.

## Fix

`commit(next?: number | null)` now accepts an optional pre-parsed value. `NumberFieldControl` passes `parsed` directly:

```ts
const parsed = root.parse(text.value)
root.value.value = parsed
root.commit(parsed)  // bypasses stale re-read
```

When `next` is omitted (all existing step/arrow callers) the behaviour is unchanged.

## Tests

- Tightened two existing tests: changed `expect([42, 5]).toContain(model.value)` → `expect(model.value).toBe(42)` now that the value is reliably committed.
- Added two regression tests that type a value, blur/Enter, and assert the parent `v-model` is updated.

Fixes #329

### DIFF
--- a/packages/0/src/components/NumberField/NumberFieldControl.vue
+++ b/packages/0/src/components/NumberField/NumberFieldControl.vue
@@ -102,7 +102,7 @@
   function onBlur () {
     const parsed = root.parse(text.value)
     root.value.value = parsed
-    root.commit()
+    root.commit(parsed)
     root.isFocused.value = false
     syncText()
   }
@@ -128,7 +128,7 @@
       Enter: () => {
         const parsed = root.parse(text.value)
         root.value.value = parsed
-        root.commit()
+        root.commit(parsed)
       },
     }
 
--- a/packages/0/src/components/NumberField/NumberFieldRoot.vue
+++ b/packages/0/src/components/NumberField/NumberFieldRoot.vue
@@ -213,7 +213,7 @@
     wrap,
     locale,
     format: formatOptions,
-    clamp: shouldClamp,
+    clamp: shouldClamp = true,
     rules = [],
     validateOn = 'blur',
     error = false,
--- a/packages/0/src/components/NumberField/index.test.ts
+++ b/packages/0/src/components/NumberField/index.test.ts
@@ -1563,9 +1563,7 @@ describe('numberField', () => {
       await wait()
       input.dispatchEvent(new FocusEvent('blur'))
       await wait()
-      // Either the v-model committed via onBlur, or it's still 5;
-      // cover the path itself, not the v-model bridge.
-      expect([42, 5]).toContain(model.value)
+      expect(model.value).toBe(42)
     })
 
     it('should commit on Enter key', async () => {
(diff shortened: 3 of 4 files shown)


johnleider:
Thanks for this, @sridhar-3009. This same fix landed on `master` via #371 — passing the parsed value directly to `commit()` so typed values propagate to the parent `v-model` on blur and Enter, avoiding the stale same-tick model re-read. The #329 regression tests (blur + Enter) are in as well, so #344 is now a duplicate. Closing as superseded — appreciate the contribution.

PR #372, opened 2026-06-25
feat(locale): add ti() for inline component aria defaults (WCAG 4.1.2)

### LINKED ISSUE #196: locale.t() needs a key-level fallback — WCAG 4.1.2 gap

sridhar-3009:
Gives every v0 component a built-in English accessible name when no locale plugin is installed (WCAG 4.1.2), **without bundling any locale strings into the runtime**.

## Approach

- Adds `locale.ti(key, ...params): string | undefined` ("translate if exists") — same lookup as `t()` (selected locale → fallback locale → interpolation) but returns `undefined` on a miss instead of echoing the key.
- Components render their aria labels as `locale.ti(key, ...params) ?? '<inline English default>'`, so the default lives at the call site and tree-shakes with the component.
- `@vuetify/v0/locale/messages/en` is exposed as an **optional** export (the canonical English catalog) for seeding translations or registering full English coverage — it is never imported by the runtime fallback.
- ESLint gate: a `locale.ti()` call inside a component must be `??`-guarded, so a component can't ship an undefined accessible name.

Button is intentionally left for #373 (which rewrites the same `aria-label` line).

Closes #196.


johnleider:
Hey @sridhar-3009 — thanks for tackling the WCAG aria-label gap. I pushed directly to your branch (maintainer edit) to steer the approach, since it lands a public-API decision. Quick rundown of what changed and why.

**What changed**
- Dropped the bundled English fallback: `createLocaleFallback()` no longer imports `messages/en`, and the `onMissing` hook was removed.
- Added `locale.ti(key, ...params): string | undefined` ("translate if exists") — same lookup as `t()`, but it returns `undefined` on a miss instead of echoing the key back.
- Every component now sets its accessible name as `locale.ti(key, ...params) ?? '<inline English default>'`, so the default lives at the call site.
- `@vuetify/v0/locale/messages/en` stays as an **optional** export (and I wired the missing `tsdown` build entry so it actually ships) — it's just never imported by the runtime anymore.
- Added an ESLint rule so a component `locale.ti()` must be `??`-guarded, plus per-component tests covering both branches (inline default when unconfigured, translation when a locale is registered).

**Why**
The bundled fallback meant every app shipped the entire English string table even if it rendered a single component. Keeping the default at the call site tree-shakes it with the component, keeps `messages/en` opt-in for anyone seeding translations, and still gives unconfigured apps a real accessible name — the original WCAG 4.1.2 goal.

Button is intentionally untouched here; it's handled in #373, which rewrites the same `aria-label` line. All green. Thanks again for getting this moving!

PR #373, opened 2026-06-25
fix(Button): suppress auto aria-label in renderless mode (WCAG 4.1.2)

### LINKED ISSUE #330: Button: raw locale key 'Button.label' leaks as accessible name when token unregistered (overrides visible text)
### Observed
In an app with no `Button.label` locale token registered (the framework builder), screen readers announce **"Button.label"** — the raw locale key — as the accessible name on Button.Root usages across the app, per the Chromium accessibility tree. Affected examples include icon-only buttons (remove-token X) and, notably, buttons with visible text ("Start configuring", "Save & Next", "Open in Playground"), where the bogus aria-label overrides the visible text name.

### Mechanism
1. `ButtonRoot.vue:217`:
   ```ts
   'aria-label': ariaLabel || (isSolo.value ? locale.t('Button.label') : undefined),
   ```
2. `useLocale/adapters/v0.ts:46` — `t()` returns the key verbatim when unregistered, so the fallback is the literal string `Button.label`.
3. No locale bundle in the repo defines `Button.label`, so every consumer that doesn't register it ships the raw key.

This is the same class of gap as #196 (components should coalesce to an English default when `t()` returns the key verbatim — PHILOSOPHY §5.5) but on the aria-label path. Additionally, `isSolo` appears to be true for renderless usages that clearly have text content, which makes the override land on text buttons too — that detection deserves its own look while in there.

### Proposed fix
Two layers, consistent with §5.5:
1. Never emit a raw key: coalesce `locale.t('Button.label')` to an English default (or emit no aria-label at all) when the lookup returns the key verbatim.
2. Don't set a fallback aria-label when the button has visible text content — per WCAG, aria-label should not override a usable visible name. Restrict the fallback to genuinely icon-only buttons, and verify why `isSolo` is true for renderless text-bearing usages.

Regression tests next to the existing aria coverage at `packages/0/src/components/Button/index.test.ts:602-627`:
- solo button without explicit `ariaLabel` must not expose `aria-label="Button.label"`
- button with visible text must not get any default aria-label

sridhar-3009:
Fixes #330

> Note: The raw-key leak (`'Button.label'` instead of `'Button'`) is addressed separately in #372. This PR fixes the `isSolo` / renderless mismatch that causes the label to override visible text at all.

## Problem

In renderless mode (`renderless: true`), consumers own the full DOM — they write the button element and its content themselves. If they include `Button.Icon` in their rendering (to get icon slot-props), `Button.Icon.onMounted` detects `root.single.size === 0` (no `Button.Content` registered) and sets `isSolo = true`. That causes `ButtonRoot` to emit an `aria-label`, overriding the consumer's own visible text.

## Fix

One-line guard in `ButtonRoot.vue`:

```diff
- 'aria-label': ariaLabel || (isSolo.value ? locale.t('Button.label') : undefined),
+ 'aria-label': ariaLabel || (!renderless && isSolo.value ? locale.t('Button.label') : undefined),
```

In renderless mode the consumer is responsible for the accessible name — the component stays out of the way.

## Tests

Two new regression tests:
1. `renderless` + `Button.Icon` + no explicit `ariaLabel` → no `aria-label` attribute (was incorrectly emitted before)
2. Non-renderless + `Button.Content` (text) + no `Button.Icon` → no `aria-label` attribute (sanity check that text buttons never get a default label)

## Checklist
- [x] `pnpm test:run` — 6249 tests pass
- [x] `pnpm typecheck` — clean
- [x] `pnpm lint:fix` — clean
- [x] `pnpm repo:check` — clean

### DIFF
--- a/.changeset/button-aria-label-330.md
+++ b/.changeset/button-aria-label-330.md
@@ -1,0 +1,5 @@
+---
+"@vuetify/v0": patch
+---
+
+fix(Button): don't auto-set aria-label in renderless mode — in renderless mode the consumer owns the DOM and is responsible for the accessible name; the automatic icon-only fallback no longer overrides visible text in mixed-content renderless usages
--- a/knip.json
+++ b/knip.json
@@ -142,9 +142,13 @@
     "**/*.d.ts"
   ],
   "ignoreDependencies": [
+    "@changesets/cli",
     "@js-temporal/polyfill",
     "@vue/repl",
     "@vuetify/paper"
+  ],
+  "ignoreBinaries": [
+    "changeset"
   ],
   "ignoreExportsUsedInFile": {
     "interface": true,
--- a/knip.json
+++ b/knip.json
@@ -142,13 +142,9 @@
     "**/*.d.ts"
   ],
   "ignoreDependencies": [
-    "@changesets/cli",
     "@js-temporal/polyfill",
     "@vue/repl",
     "@vuetify/paper"
-  ],
-  "ignoreBinaries": [
-    "changeset"
   ],
   "ignoreExportsUsedInFile": {
     "interface": true,
--- a/packages/0/src/components/Button/ButtonRoot.vue
+++ b/packages/0/src/components/Button/ButtonRoot.vue
@@ -214,7 +214,7 @@
     'aria-disabled': isDisabled.value || isPassive.value,
     'aria-busy': isLoading.value ? true : undefined,
     'aria-pressed': group ? isSelected.value : undefined,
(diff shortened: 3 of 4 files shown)


johnleider:
Superseded by https://github.com/vuetifyjs/0/pull/390 — reopened on the org repo to fold in the `locale.ti()` migration alongside your renderless fix. Merging master into this fork branch (needed for the `ti` API) tripped a `workflow`-scope push restriction, so I couldn't push it here directly. Your renderless fix carries over intact. Thanks @sridhar-3009!

PR #374, opened 2026-06-25
docs(Dialog): document top-layer stacking limitation and Snackbar workarounds

### LINKED ISSUE #279: [Bug Report][1.0.0-beta.0] Dialog API limitations
### Environment
**Vuetify0 Version:** 1.0.0-beta.0
**Vue Version:** 3.5.35
**OS:** macOS 10.15.7 (current)

### Steps to reproduce
1. Click to "Open dialog"
2. Click to "Show snackbar (inside dialog)"

### Expected Behavior
The snackbar always appears on top of the dialog

### Actual Behavior
The snackbar appears below the dialog

### Reproduction Link
[https://github.com/pont1s/vuetify0-dialog](https://github.com/pont1s/vuetify0-dialog)

### Other comments
Since Dialog is built on top of the Dialog API, is there a plan to address at the library level (vuetify0) the set of issues that currently exist with this API? Or should fixing them be handled by the consumer when building a design system on top of vuetify0? What is vuetify0 position on these issues?

A list of such issues can be found at https://github.com/carbon-design-system/carbon/issues/22241

One such issue is the common case where a Snackbar (Toast) needs to be displayed inside a Dialog. In this case, the expectation is that the Snackbar should always appear on top of the Dialog and remain interactive.

As a side note, I expected Snackbar to be implemented via the Popover API with a `manual` type, but even then the problem is not resolved. When using a shared queue, Snackbars are rendered in their container inside `body`, and the Dialog specification requires that a Popover be nested within the dialog, otherwise it will not be interactive.

Within my own design system, I solved this problem using a `usePopoverTopmostTeleportTarget` composable based on https://github.com/whatwg/html/issues/9936#issuecomment-2682028665, though it is far from perfect.

sridhar-3009:
Closes #279

## What

Adds a **Known Limitations** section to the Dialog docs explaining why overlays (Snackbar, Tooltip, Popover) rendered outside a `<dialog>` appear below it, and shows two concrete workarounds consumers can use today.

## Why

The native `<dialog showModal()>` element promotes itself to the browser's top layer — nothing rendered in the normal document flow (including portals to `body`) can appear above it. This is a browser-level constraint, not a v0 bug. The issue author asked whether v0 would address this at the library level; the existing `Snackbar.Portal teleport` prop already provides the escape hatch — we just need to document it.

## Workarounds documented

1. **`teleport="false"`** — renders `Snackbar.Portal` inline inside `Dialog.Content`, inheriting the top-layer context
2. **Dynamic teleport target** — bind `Snackbar.Portal :teleport` to a `<div>` placed inside the dialog when it is open

## Checklist
- [x] No code changes — docs only
- [x] `pnpm test:run` — passes (no code changed)
- [x] `pnpm repo:check` — clean

### DIFF
--- a/apps/docs/src/pages/components/disclosure/dialog.md
+++ b/apps/docs/src/pages/components/disclosure/dialog.md
@@ -69,6 +69,58 @@ The `blocking` prop disables scrim-based dismissal entirely — the dialog can o
 </template>
 ```
 
+## Known Limitations
+
+### Overlays inside a modal dialog (Snackbar, Tooltip, Popover)
+
+The native `<dialog>` element with `showModal()` promotes itself to the browser's **top layer** — a rendering surface that sits above all normal document content. This has one important consequence: any overlay rendered _outside_ the dialog (e.g., a `Snackbar.Portal` teleported to `body`, or a `Tooltip` inside a portal) will appear **below** the dialog, not above it.
+
+This is a browser-level constraint, not a v0 bug. The two ways to work around it are:
+
+**Option 1 — Render the overlay inside the dialog element**
+
+Place `Snackbar.Portal` (or any overlay) as a direct child of `Dialog.Content`. The content will render inside the native `<dialog>` element, which is already in the top layer:
+
+```vue
+<template>
+  <Dialog.Root v-model="open">
+    <Dialog.Content>
+      <Dialog.Title>Settings</Dialog.Title>
+      <!-- ... dialog body ... -->
+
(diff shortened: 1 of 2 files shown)


johnleider:
Superseded by https://github.com/vuetifyjs/0/pull/391 — your branch predated #381's `gn-example` reformatting of `dialog.md`, so it conflicted on the page format. Re-applied the top-layer limitation section onto the current format there. Thanks @sridhar-3009!

PR #383, opened 2026-06-25
feat(createCombobox): commit free-text values in non-strict mode

johnleider:
## Summary

`createCombobox` exposed a `strict` option whose only real effect was the `aria-autocomplete` attribute. The documented "reverts query to selected value on close" behaviour was unconditional (driven by `pristine` + `display`, fired on every `close()`), so `strict` was effectively cosmetic — and the **default** non-strict combobox silently discarded any typed text that didn't match a registered option. A non-strict / free-solo combobox is supposed to accept arbitrary values.

This PR makes non-strict mean what it says: typed text that matches no option is committed as a value.

## What changed

- **`commit()` action** on the combobox context. On confirm it selects an exact value match if one exists; otherwise, when `!strict`, it mints a registry ticket for the typed text, selects it, and emits it through `v-model`. When `strict`, unmatched text is discarded.
- **`ComboboxControl`** wires `commit()` to **Enter** (when nothing is highlighted) and **Tab** (accept highlight, else commit). **Escape** still cancels/reverts — which is why commit lives in its own action rather than in `close()` (Escape and click-outside both route through `close()`).
- **Ad-hoc value hygiene:** minted free-text tickets are kept out of virtual-focus navigation (they have no rendered option / DOM node), pruned when superseded in single-select, and cleared on `clear()`.
- **Docs/JSDoc corrected:** the `strict` prop, module remarks, Strict Mode prose, the `aria-autocomplete` note, and the keyboard table previously described the unconditional revert as if it were strict-gated.

## Design note

This is the "register-on-commit" approach (vs. a parallel free-value channel) — it composes with the existing ticket/`useProxyModel` machinery instead of bolting a second value path onto every selection consumer. Single-select is the primary path; multiple-select (free-text tags) works but is the lighter-tested surface — removing a tag via direct deselect leaves an orphaned-but-deselected ticket until `clear()`.

## Tests

No automated tests are included in this commit, per the repo convention of not adding tests unless requested. The existing 119 combobox tests pass unchanged (the new behaviour is keyed on Enter/Tab; the prior "strict mode" tests exercise `close()` and are unaffected). The `commit()` path — free-text emit, strict discard, exact-match dedup, single-select pruning — is currently uncovered; happy to add a focused `describe('free-text commit')` block if you want it in this PR.


packages/0/src/composables/createCombobox/index.ts
```diff
@@ -256,6 +319,10 @@ export function createCombobox (options: ComboboxOptions = {}): ComboboxContext
     for (const id of Array.from(selection.selectedIds)) {
       selection.unselect(id)
```
johnleider: **[bug]** `clear()` was rewritten to `selection.unselect(id)` plus `unregister` of ad-hoc ids, and it dropped the existing drain. On this PR's own base (and current `dev`/`master`) `clear()` was: skip when the combobox is disabled; delete from `selectedIds` directly so disabled *tickets* still empty out; stop at one remaining id when `mandatory`. `createSelection.unselect()` no-ops when the context or ticket is disabled, and `drain()` refuses to drop the last mandatory id. Then `selection.unregister(id)` on every ad-hoc ticket still runs (`createModel.unregister` does `selectedIds.delete(id)`), so a mandatory combobox whose only selection is a committed free-text value ends up with an empty selection, while a disabled selected option is left selected. The new ad-hoc unregister is needed; the unselect loop is a regression.

**Suggestion:** Restore the previous drain (`if (toValue(disabled)) return`, then `selectedIds.delete` with the mandatory floor) and only then unregister ad-hoc tickets that are no longer selected. Do not route wholesale clear through `unselect()`. Add tests for `disabled`, `mandatory`, and a disabled selected ticket.

packages/0/src/components/Combobox/ComboboxControl.vue
```diff
@@ -108,6 +110,15 @@
           break
         }
         case 'Tab': {
+          const highlighted = context.cursor.highlightedId.value
+          if (isUndefined(highlighted)) {
+            context.commit()
+          } else {
+            context.select(highlighted)
```
johnleider: **[bug]** Tab with a highlighted option calls `context.select(highlighted)`. In multiple mode `select()` is `selection.toggle()`, so tabbing away from an already-selected highlighted item *deselects* it, then the extra `close()` hides the list. The keyboard table says Tab "Accept the highlighted item … then move focus" — accept, not toggle. Enter toggling the highlight is the right listbox gesture; Tab is a leave/confirm gesture and was previously close-only, so this also changes Tab for every combobox, not just free-text.

**Suggestion:** On Tab, select without toggling (e.g. `selection.select(id)` or a confirm helper that no-ops if already selected), then `close()`. Leave Enter on `context.select()` so multiple-select Enter still toggles. Cover "Tab with a selected highlight in `multiple` does not unselect".

packages/0/src/composables/createCombobox/index.ts
```diff
@@ -196,7 +202,9 @@ export function createCombobox (options: ComboboxOptions = {}): ComboboxContext
 
   const cursor = useVirtualFocus(
     () => selection.values()
-      .filter(ticket => filtered.value.has(ticket.id) && !toValue(ticket.disabled))
+      // Ad-hoc free-text values have no rendered option, so keep them out of
+      // keyboard navigation — they're committed values, not list options.
+      .filter(ticket => filtered.value.has(ticket.id) && !toValue(ticket.disabled) && !adhoc.has(ticket.id))
```
johnleider: **[bug]** Ad-hoc tickets are excluded from virtual-focus, but `items` is still `[...selection.values()]`, so they remain in `filtered`. `ClientComboboxAdapter` sets `isEmpty` from `filtered.size === 0`, and `ComboboxEmpty` renders on `context.isEmpty`. After committing a unique string (e.g. `"Durian"`) and typing it again, `filtered` is `{adhocId}`, `isEmpty` is false, every `ComboboxItem` is `v-show` hidden, and the list is a blank box with no empty state. The unit test even asserts `filtered.value.has(adhocId) === true`.

**Suggestion:** Exclude `adhoc` from the adapter `items` computed (same predicate as the cursor filter), so filtering, `isEmpty`, and empty-state UI only see real options. Keep ad-hoc ids on `selection` for `selectedValues` / v-model. Flip the unit test to expect `filtered` not to contain the ad-hoc id.

packages/0/src/components/Combobox/ComboboxControl.vue
```diff
@@ -108,6 +110,15 @@
...
+          }
+          // Multiple-select commit/select keeps the menu open; close it here so the
+          // listbox doesn't float over the next control (useClickOutside dismisses on
+          // pointer events, not on focus leaving via Tab).
           context.close()
```
johnleider: **[bug]** Tab commits (or accepts the highlight) then `close()`s, because — per the comment — `useClickOutside` only dismisses on pointer events. That outside handler still calls `close()`, which always blanks `query` and sets `pristine`, so the input reverts to the previous selection. Typing a free-text value and clicking away, or clicking a form Submit, therefore drops the value that Tab would have written to v-model. The component FAQ says a non-strict combobox "accepts unmatched text as a free-text value through v-model" on confirm; pointer leave is the other leave-field path and currently cancels.

**Suggestion:** Either call `commit()` from the click-outside path (and from any other dismiss that is "leave field" rather than Escape), or document pointer dismiss as cancel alongside Escape. If click-outside stays cancel, the keyboard table / FAQ must say so — otherwise the default non-strict combobox silently loses typed text on the most common confirm-by-mouse path.

packages/0/src/composables/createCombobox/index.ts
```diff
@@ -12,7 +12,7 @@
  * - Adapter-based filtering (client-side or server-side via ClientComboboxAdapter/ServerComboboxAdapter)
  * - Single or multi-select mode
  * - Virtual focus keyboard navigation (aria-activedescendant pattern)
- * - Strict mode: reverts query to selected value on close if no match
+ * - Strict mode: constrains values to registered options; non-strict commits typed text as a new value
```
johnleider: **[bug]** Module JSDoc now says non-strict commits typed text, and the component page/FAQ/keyboard table were updated to match. `apps/docs/src/pages/composables/forms/create-combobox.md` was not: `close()` is still "applies strict revert if needed", `strict` is still "Revert query on close if no match", the Strict Mode recipe and FAQ still say non-strict "leaves whatever the user typed in place", and `commit()` is absent from the context table. The CountryAutocomplete example still only `select`s on Enter when `highlightedId` is set — no `commit()` — so the composable example does not demonstrate the new public API.

**Suggestion:** Update the composable page in lockstep with the component page: document `commit()`, describe `strict` as constrain-on-confirm, and wire Enter (no highlight) / Tab to `commit()` in the example. Drop the revert-on-close wording; `close()` already always clears `query` and relies on `pristine` + `display`.

.changeset/combobox-free-text.md
```diff
@@ -0,0 +1,7 @@
+---
+"@vuetify/v0": minor
+---
+
+feat(createCombobox): commit free-text values in non-strict mode
+
+A non-strict combobox (the default) now accepts typed text that matches no registered option. Confirming with **Enter** or **Tab** mints a ticket for the typed value, selects it, and emits it through `v-model`; `strict` keeps the constrained behaviour and discards unmatched text on confirm. Adds a `commit()` action to the combobox context. Free-text values are kept out of keyboard navigation and pruned when superseded (single-select) or cleared.
```
johnleider: **[suggestion]** Minor bump of `@vuetify/v0` on `dev` is the right train for a new public `commit()` action. The body then explains the internal ad-hoc Set (keyboard-nav exclusion, single-select prune, clear). The releasing skill keeps changelog copy on the consumer-visible delta, not the mechanism.

**Suggestion:** Keep the Enter/Tab vs `strict` behavior and that `commit()` is now on the context. Drop the pruning / virtual-focus sentence; that belongs in the PR description.

Issue #419, opened 2026-06-29
Overflow "responsive toolbar" docs example: popover content is always visible (renders open without clicking)


johnleider:
### Where
Overflow docs page → **Responsive toolbar** example.
- Page: `/components/semantic/overflow`
- Source: `apps/docs/src/examples/components/overflow/OverflowToolbar.vue`

### Symptom
The `+N more` overflow popover **renders open on load** — its content is visible without the activator ever being clicked.

### Repro
1. Open the Overflow page.
2. Narrow the viewport until the toolbar overflows (≤ ~500px wide).
3. The `+6 more` popover menu (Export / Print / Rename / Duplicate / Archive / Delete) is shown without interacting with the activator.

### Evidence (DOM, at 500px wide)
- `Popover.Content` (`.min-w-40`) is `display: flex`, 160×212, **visible**.
- The `+6 more` activator has `aria-expanded = null` and `popovertarget = "v-…"` (native Popover API) — i.e. the content is **not starting in the closed state**.

### Expected
Popover content stays hidden until the activator is clicked.

### Likely cause
`Popover.Content` isn't gated on the open state in this usage (the native popover isn't starting hidden). Looks component-level (`C: Popover`) surfaced via this example, rather than purely an example mistake — triage to confirm.

### Note
Found while reviewing #417's docs preview. **Not** caused by #417 — that PR only refactors `OverflowRoot.vue` item-visibility caching and doesn't touch Popover/open-state. Pre-existing.

Issue #408, opened 2026-06-29
[Chore] @vuetify/v0/browser subpath ships no type declarations (attw UntypedResolution)


johnleider:
## Summary

`are-the-types-wrong` flags the `@vuetify/v0/browser` export with `UntypedResolution` ("No types"): `./browser` maps to `./dist/browser/index.js`, but `dist/browser/` ships only `index.js` — no `.d.ts`. A TypeScript consumer importing `@vuetify/v0/browser` gets no types.

## Location

- `packages/0/package.json` exports: `"./browser": "./dist/browser/index.js"` (no `types` condition).
- `dist/browser/` contains `index.js` only.
- Surfaced by the new `repo:exports` CI check (`publint` + `attw`), which currently baselines this via `attw --pack ./packages/0 --profile esm-only --exclude-entrypoints browser` so the guard lands green while still catching new untyped/broken exports on every other entrypoint.

## Decision needed

1. Emit a `.d.ts` for the browser entry + add a `types` condition to the `./browser` export, then drop `--exclude-entrypoints browser` from `repo:exports`; **or**
2. Accept it as intentional (the browser bundle is for `<script>`-tag use, not TS imports) and keep the documented exclusion.

Low priority — the browser entry is primarily for direct-browser use. Tracked so the baseline exclusion is a conscious decision, not a silent gap.

Issue #400, opened 2026-06-29
[Bug Report] createTokens: resolve() returns inherited prototype members for unknown path segments


johnleider:
## Summary

`createTokens` `resolve()` walks alias path segments with `segment in current`, which traverses the prototype chain. When a prefix resolves to a plain-object token value, resolving a path whose final segment names an `Object.prototype` member (`constructor`, `toString`, `hasOwnProperty`, `__proto__`, …) returns that inherited member instead of `undefined` + the "Path not found" warning.

## Location

- `packages/0/src/composables/createTokens/index.ts:234-243` — `resolve()` uses `if (!isObject(current) || !(segment in current))`.
- Contrast its twin `flatten()` (`:419-420`, `:437-438`), which already guards with `Object.prototype.hasOwnProperty.call(...)` + `UNSAFE_KEYS`. `UNSAFE_KEYS` is imported (`:36`) but used only in `flatten`.
- Reachable when a whole object is registered as a token value (flat mode, `:455-457`) or a `{ $value: {...} }` alias is unwrapped (`:232` / `:242`).

## Impact

This is a **read** of an inherited member, not a write — **not** prototype pollution, no info disclosure beyond global builtins, no attacker surface (token config and the `resolve()` argument are developer-authored; `createTokens` backs `useTheme` / `useLocale` / `useFeatures`). The real-world bite is a correctness inconsistency: `resolve()` can hand back a builtin function instead of `undefined`. Low severity — consistency / defense-in-depth.

## Suggested fix

Mirror `flatten`'s guard in the `resolve` loop (`:235`):

```ts
if (!isObject(current) || UNSAFE_KEYS.has(segment) || !Object.prototype.hasOwnProperty.call(current, segment))
```

Add tests asserting `resolve('{obj.constructor}')` / `resolve('{obj.__proto__}')` return `undefined`. Since `flatten` can never register an unsafe/inherited key, mirroring the guard carries zero false-negative risk. Describe it in the changelog as "resolve() no longer returns inherited prototype members," not as a security fix.

PR #429, opened 2026-06-30
fix(createModel): disabled guard missing from apply() path (#398)

### LINKED ISSUE #398: [Bug Report] createSelection: disabled items selectable via multiple-mode v-model
## Summary

In multiple-selection mode, applying a value through `v-model` (the `apply()` path) can select a **disabled** item, even though `select()`, `unselect()`, and `toggle()` all correctly reject disabled items. This violates the documented contract that a disabled model/item makes all selection operations no-ops.

## Location

- `packages/0/src/composables/createModel/index.ts:434` — `apply()` browse-fallback adds the resolved id directly (`selectedIds.add(id)`) with no disabled check.
- `packages/0/src/composables/createSelection/index.ts:260` — `apply()` multiple branch adds directly (`model.selectedIds.add(id)`) with no disabled check.
- Contrast: `createModel.select()` (`:385-393`) guards both instance-level and per-ticket `disabled`; `createSelection.apply()` **single** branch (`:266`) routes through `model.select(next)` and is already guarded.
- Documented contract: `createModel/index.ts:355-360` JSDoc — "Disabled model — all selection operations are no-ops".

## Impact

Affects multiple-mode `createSelection` / `createGroup` / `createNested` via `useProxyModel` (multi-`Select`, `Combobox`, `CheckboxGroup`, `Treeview`, multi-`ExpansionPanel`). Pushing a disabled item's value into the bound array (reset-to-default, server-restored state) renders the disabled item as selected. `useProxyModel` does not self-correct — once the id is in `selectedIds`, its value is reported back. Non-destructive state inconsistency, not a crash; no attacker surface (consumer owns the `v-model`).

## Suggested fix

Guard only the **add** paths; do **not** blanket-route through `select()`:

- `createSelection/index.ts:260` — add an inline per-ticket guard (resolve `model.get(id)`, skip if `toValue(ticket.disabled)`). ⚠️ Do **not** replace with `model.select(id)`: `select` clears based on the model's own `multiple` (false in single-mode-with-override), which would break the existing `apply(['A','B'], { multiple: true })` override test (`createSelection/index.test.ts:997-1008`).
- `createModel/index.ts:434` — routing through `select(id)` is safe here.
- Leave the ref-write value-sync path (`createModel:419-425`) unguarded — it writes into already-selected tickets and must keep syncing.
- Add a disabled + apply regression test (none currently exists).

## Open question (decide before fixing)

Should `v-model` **sync** honor instance-level `disabled` at all, or does `disabled` gate only user interaction (so programmatic/restored state should mirror regardless)? The per-ticket guard is clearly correct either way; the instance-level question affects whether `useProxyModel` should revert a consumer's programmatic write.

sridhar-3009:
Fixes #398

## Root cause

`apply()` uses the browse-resolution fallback to select an item by value. Unlike `select()` and `toggle()`, this path had no check for `disabled` at either the model instance or ticket level, so `v-model` could silently select a disabled item.

## Fix

After finding the ticket via `registry.browse()`, mirror the guards from `select()`:

```ts
if (!toValue(disabled) && ticket && !toValue(ticket.disabled)) {
  selectedIds.add(id)
}
```

## Tests added

- `should not select a disabled ticket via apply`
- `should not select when model instance is disabled via apply`

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -22,13 +22,12 @@
         "src/pages/**/*.{vue,md}",
         "build/**/*.ts",
         "vite.config.*",
-        "src/examples/**/*.vue"
+        "src/examples/**/*.vue",
+        "src/examples/**/*.ts"
       ],
       "ignore": [
         "src/typed-router.d.ts",
-        "src/examples/guide/building-frameworks/my-ui/src/**/*.ts",
         "src/examples/guide/building-frameworks/my-ui/vite.config.ts",
-        "src/examples/composables/create-trinity/toasts.ts",
         "src/skillz/tours/**/index.ts"
       ],
       "paths": {
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "preview:docs": "pnpm --filter=docs preview",
     "repo:knip": "knip",
     "repo:sherif": "sherif",
-    "repo:exports": "publint ./packages/0 && attw --pack ./packages/0 --profile esm-only --exclude-entrypoints browser",
+    "repo:exports": "publint ./packages/0 && attw --pack ./packages/0 --profile esm-only --exclude-entrypoints browser --exclude-entrypoints style.css",
     "repo:check": "pnpm repo:knip && pnpm repo:sherif",
     "validate": "pnpm lint && pnpm typecheck && pnpm test:run",
     "changeset": "changeset",
--- a/packages/0/src/components/Popover/PopoverContent.vue
+++ b/packages/0/src/components/Popover/PopoverContent.vue
@@ -111,3 +111,14 @@
     <slot v-bind="slotProps" />
   </Atom>
 </template>
+
+<style>
+/*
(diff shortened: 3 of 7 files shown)


sridhar-3009:
Note: this PR also includes the fixes for #400 (createTokens prototype chain) and #419 (PopoverContent CSS hiding). Happy to split into separate PRs if preferred.

johnleider:
Thanks @sridhar-3009 — and appreciate you working through the v1.0.0 backlog! This overlaps #443, which is already open for the same issue (#398) and is the version we're landing. Closing as a duplicate to consolidate on that PR. Sorry for the parallel effort, and thank you for the contribution. 🙏

johnleider:
Correction to my note above: the canonical PR for #398 is #426 — I mistyped #443 (that's a different issue, #408). The close stands; consolidating on #426. Thanks again @sridhar-3009!

PR #434, opened 2026-06-30
fix(hooks): use check-only lint in pre-push hook

### LINKED ISSUE #402: [Chore] pre-push hook runs lint:fix, mutating the tree and shipping unfixed commits
## Summary

The `pre-push` hook runs `pnpm lint:fix`. `eslint --fix` exits 0 after auto-fixing, so the push proceeds while the fixes land as **unstaged** working-tree edits — the already-committed unfixed code is pushed, and CI's check-only `lint` then fails on a commit that "passed" locally.

## Location

- `package.json:46-47` — `"simple-git-hooks": { "pre-push": "pnpm lint:fix && pnpm typecheck && pnpm test:run && pnpm repo:check" }` (mirrored in `.git/hooks/pre-push`).
- CI runs check-only `pnpm lint` (`.github/workflows/pr-checks.yml`), so the two diverge.

## Impact

Contributor DX only (no consumer or attacker surface), but real and recurring: a green local push can produce red CI plus a surprise dirty working tree.

## Suggested fix

Change the `pre-push` command from `pnpm lint:fix` to check-only `pnpm lint` so local pre-push mirrors CI exactly — a lint error blocks the push loudly instead of silently rewriting the tree. Prefer this over moving a blanket `lint:fix` to `pre-commit` (a full-repo fix on every commit is heavy); if a pre-commit fix is wanted, scope it with lint-staged.

sridhar-3009:
## Summary
- `pre-push` was running `pnpm lint:fix` which auto-fixes files as unstaged edits and then exits 0
- The already-committed (unfixed) code gets pushed, then CI's check-only `pnpm lint` fails on it
- Changed to `pnpm lint` (check-only) so any lint error blocks the push instead of silently fixing it in the working tree

Closes #402

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "prepare": "simple-git-hooks"
   },
   "simple-git-hooks": {
-    "pre-push": "pnpm lint:fix && pnpm typecheck && pnpm test:run && pnpm repo:check"
+    "pre-push": "pnpm lint && pnpm typecheck && pnpm test:run && pnpm repo:check"
   },
   "devDependencies": {
     "@arethetypeswrong/cli": "catalog:",


johnleider:
Thanks @sridhar-3009 — and appreciate you working the v1.0.0 backlog! This overlaps #430, already open for the same issue (#402) and the version we're landing. Closing as a duplicate to consolidate on #430. Sorry for the parallel effort, and thanks for the contribution. 🙏

PR #435, opened 2026-06-30
fix(build): correct tsdown alias paths to packages/0/src

### LINKED ISSUE #403: [Chore] tsdown config aliases resolve outside packages/0/src (also packages/paper)
## Summary

The `@` and `#v0` bundler aliases in `packages/0/tsdown.config.mts` resolve to non-existent directories. The build is green only because the keys never match real imports — a latent footgun.

## Location

- `packages/0/tsdown.config.mts:7-8`:
  - `new URL('../src', import.meta.url)` → `…/packages/src` (does not exist; the real tree is `packages/0/src`)
  - `new URL('../../0/src', import.meta.url)` → `…/0/src` (does not exist)
- Used as `alias: { '@': at, '#v0': v0 }` (`:27-30`, `:56-59`).
- The build works because `@` is unused in source and `#v0/...` resolves via the package.json `imports` field, so the alias never wins. If a `@/...` import were added, it would resolve to the nonexistent `packages/src` and fail confusingly.
- Correct precedent: `packages/0/vitest.config.ts:9,13` and `packages/genesis/tsdown.config.mts:7` (single `../0/src`).
- **Same defect** in `packages/paper/tsdown.config.mts:8-9`.

## Suggested fix

Delete the `at` / `v0` consts and both `alias` blocks (`@` is unused; `#v0` already resolves via package.json `imports` + tsconfig paths — the build is proven to work without them). If keeping the aliases, correct them to `new URL('src', import.meta.url)`. Fix `packages/paper` in the same pass (bug-family).

sridhar-3009:
## Summary
- `new URL('../src', import.meta.url)` from `packages/0/` resolves to `packages/src` (non-existent)
- `new URL('../../0/src', import.meta.url)` resolves to a path outside the monorepo root
- Both are corrected to `new URL('./src', import.meta.url)` which resolves to `packages/0/src`
- Build was green only because `@` is unused in source and `#v0` is resolved via `package.json#imports` before hitting the alias

Closes #403

### DIFF
--- a/packages/0/tsdown.config.mts
+++ b/packages/0/tsdown.config.mts
@@ -4,8 +4,8 @@ import Vue from 'unplugin-vue/rolldown'
 
 import pkg from './package.json' with { type: 'json' }
 
-const at = fileURLToPath(new URL('./src', import.meta.url))
-const v0 = fileURLToPath(new URL('./src', import.meta.url))
+const at = fileURLToPath(new URL('src', import.meta.url))
+const v0 = fileURLToPath(new URL('src', import.meta.url))
 const __VERSION__ = JSON.stringify(pkg.version)
 
 export default defineConfig([{
--- a/packages/0/tsdown.config.mts
+++ b/packages/0/tsdown.config.mts
@@ -4,8 +4,8 @@ import Vue from 'unplugin-vue/rolldown'
 
 import pkg from './package.json' with { type: 'json' }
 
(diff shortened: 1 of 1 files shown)


johnleider:
Thanks @sridhar-3009 — and appreciate you working the v1.0.0 backlog! This overlaps #431, already open for the same issue (#403) and the version we're landing. Closing as a duplicate to consolidate on #431. Sorry for the parallel effort, and thanks for the contribution. 🙏

PR #436, opened 2026-06-30
fix(exports): add type declarations to browser subpath

### LINKED ISSUE #408: [Chore] @vuetify/v0/browser subpath ships no type declarations (attw UntypedResolution)

sridhar-3009:
## Summary
- `@vuetify/v0/browser` had no `types` condition — `are-the-types-wrong` flagged it as `UntypedResolution`
- Added `"types": "./dist/index.d.mts"` to both the dev `exports["./browser"]` and `publishConfig.exports["./browser"]`
- Updated `tsdown.config.mts` `customExports` to emit the same condition on publish
- Removed the `--exclude-entrypoints browser` baseline from `repo:exports` script now that types resolve correctly

Closes #408

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "preview:docs": "pnpm --filter=docs preview",
     "repo:knip": "knip",
     "repo:sherif": "sherif",
-    "repo:exports": "publint ./packages/0 && attw --pack ./packages/0 --profile esm-only --exclude-entrypoints browser",
+    "repo:exports": "publint ./packages/0 && attw --pack ./packages/0 --profile esm-only",
     "repo:check": "pnpm repo:knip && pnpm repo:sherif",
     "validate": "pnpm lint && pnpm typecheck && pnpm test:run",
     "changeset": "changeset",
--- a/packages/0/package.json
+++ b/packages/0/package.json
@@ -344,8 +344,8 @@
     },
     "./package.json": "./package.json",
     "./browser": {
-      "development": "./src/index.ts",
       "types": "./dist/index.d.mts",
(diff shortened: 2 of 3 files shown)


johnleider:
Thanks @sridhar-3009 — and appreciate you working the v1.0.0 backlog! This overlaps #443, already open for the same issue (#408) and the version we're landing. Closing as a duplicate to consolidate on #443. Sorry for the parallel effort, and thanks for the contribution. 🙏

PR #437, opened 2026-06-30
fix(useTheme): validate adapter prefix against SAFE_IDENT

### LINKED ISSUE #405: [Bug Report] useTheme: theme prefix interpolated into CSS without SAFE_IDENT validation
## Summary

`useTheme`'s CSS generator sanitizes theme names, color keys, and values, but interpolates the adapter `prefix` into the generated CSS **without validation** — a malformed prefix can break out of the declaration block and inject arbitrary CSS rules.

## Location

`packages/0/src/composables/useTheme/adapters/adapter.ts`:
- `:16` `UNSAFE_CSS = /url\s*\(|@import|expression\s*\(|[{}<>]/i`, `:17` `SAFE_IDENT = /^[a-zA-Z0-9_-]+$/`
- `:42` theme name guarded with `SAFE_IDENT`; `:45` keys guarded with `SAFE_IDENT`, values with `!UNSAFE_CSS`
- `:46` / `:49` `--${this.prefix}-${key}` and `var(--${this.prefix}-on-background)` — **prefix interpolated raw**
- `:28-29` constructor stores `prefix` unvalidated

The repo's own `.claude/rules/implementation.md` "Security primitives" table already prescribes this exact guard for CSS interpolation ("validate keys with `SAFE_IDENT`, reject values matching `UNSAFE_CSS`"). The prefix is the one input in `generate()` that skips it.

## Impact

Low in core v0 — `prefix` is developer-supplied config (constructor arg, has a default). It becomes a real injection vector only if a downstream design system pipes user/tenant-controlled input into the theme prefix. Unlike a `querySelector` selector (where escaping would break valid input), the prefix is interpolated into a generated CSS string injected via `adoptedStyleSheets` / `<style>`, so a `}` breakout genuinely injects rules (UI redress, attribute-selector exfiltration). The fix is non-breaking (valid prefixes already match `SAFE_IDENT`) and matches the sanitization already applied to adjacent inputs in the same function.

## Suggested fix

Validate `this.prefix` against `ThemeAdapter.SAFE_IDENT` in the constructor (throw, or fall back to the default) — mirroring the existing key/value guards. One line, no behavior change for valid prefixes.

sridhar-3009:
## Summary
- Theme adapter `prefix` was interpolated into CSS as `--${prefix}-${key}` without validation
- A crafted prefix like `bad}{...` could break out of the declaration block and inject arbitrary CSS rules
- Added SAFE_IDENT check in `ThemeAdapter` constructor — throws on any prefix that doesn't match `/^[a-zA-Z0-9_-]+$/`
- Added test covering invalid prefix shapes (`bad}{prefix`, `bad<script>`, empty string)

Closes #405

### DIFF
--- a/packages/0/src/composables/useTheme/adapters/adapter.ts
+++ b/packages/0/src/composables/useTheme/adapters/adapter.ts
@@ -26,6 +26,9 @@ export abstract class ThemeAdapter {
   dispose?: () => void
 
   constructor (prefix: string) {
+    if (!ThemeAdapter.SAFE_IDENT.test(prefix)) {
+      throw new Error(`Invalid theme prefix "${prefix}": must match /^[a-zA-Z0-9_-]+$/.`)
+    }
     this.prefix = prefix
   }
 
--- a/packages/0/src/composables/useTheme/adapters/adapter.ts
+++ b/packages/0/src/composables/useTheme/adapters/adapter.ts
@@ -1,5 +1,5 @@
 // Utilities
-import { hexToRgb, isUndefined } from '#v0/utilities'
+import { hexToRgb, isUndefined, V0Error } from '#v0/utilities'
 
 // Types
 import type { ID } from '#v0/types'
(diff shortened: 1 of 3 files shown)


johnleider:
Thanks @sridhar-3009 — and appreciate you working the v1.0.0 backlog! This overlaps #440, already open for the same issue (#405) and the version we're landing. Closing as a duplicate to consolidate on #440. Sorry for the parallel effort, and thanks for the contribution. 🙏

PR #439, opened 2026-06-30
fix(useStack): avoid module-scoped fallback singleton in SSR

### LINKED ISSUE #407: [Bug Report] useStack: module-scoped fallback leaks across SSR requests
## Summary

`useStack` falls back to a **module-scoped singleton** (`fallbackStack`) when no provider/context exists. In SSR (long-lived Node process), overlay tickets registered into this global persist across requests and aren't cleaned up — causing cross-request state bleed (wrong z-index) and unbounded memory growth.

## Location

- `packages/0/src/composables/useStack/index.ts:359-365` — `let fallbackStack` + lazy `getStackFallback()`.
- `:408-410` — `useStack()` returns the fallback when `!instanceExists()` (and on a `useContext` miss).
- `:276` cleanup is `onScopeDispose`, which is unreliable server-side.
- Documented mitigation already present (`:375`): "For SSR safety, use createStackPlugin to provide isolated context per app."

## Impact

Partial / low. Requires SSR + an overlay component (`Dialog` / `Popover` / `Snackbar`) used **without** `createStackPlugin` — which the JSDoc already flags as the wrong SSR setup. Note the "cross-request data leakage" framing (from the source review) is overstated: these are z-index tickets, not user data — it's state bleed + memory growth, not data exfiltration. Aligns with the composable-bar's SSR-hardening priority.

## Suggested fix

Guard the fallback under SSR: either throw when no context is provided server-side, or return an ephemeral per-call `createStack()` (no shared global) on the server, steering consumers to `createStackPlugin` for coordinated SSR z-index. Add an SSR test.

sridhar-3009:
## Summary
- `getStackFallback()` lazily initialized a module-scoped `fallbackStack` singleton
- In SSR (long-lived Node process) tickets registered into this global persist across requests — wrong z-index for unrelated users, unbounded memory growth
- In browser the singleton is fine (one app per tab); in SSR each call should get an isolated context so request A's overlays don't bleed into request B
- Fix: return a fresh `createStack()` when `IN_BROWSER` is false instead of the cached singleton
- Adds a test that verifies two consecutive `useStack()` calls in SSR return distinct contexts

Closes #407

### DIFF
--- a/packages/0/src/composables/useStack/index.test.ts
+++ b/packages/0/src/composables/useStack/index.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it, vi } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 import {
   createStack,
@@ -13,6 +13,14 @@ import { defineComponent, nextTick, ref, watch } from 'vue'
 
 // Types
 import type { StackContext } from '.'
+
+const mockInBrowser = vi.hoisted(() => ({ value: true }))
+
+vi.mock('#v0/constants/globals', () => ({
+  get IN_BROWSER () {
+    return mockInBrowser.value
+  },
+}))
 
 describe('createStack', () => {
   describe('useStack', () => {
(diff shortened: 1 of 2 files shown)


johnleider:
Thanks @sridhar-3009 — and appreciate you working the v1.0.0 backlog! This overlaps #442, already open for the same issue (#407) and the version we're landing. Closing as a duplicate to consolidate on #442. Sorry for the parallel effort, and thanks for the contribution. 🙏

PR #446, opened 2026-06-30
fix(ci): trigger docs/playground builds when packages change

### LINKED ISSUE #401: [CI] package-only PRs skip docs/playground build checks
## Summary

A PR that touches only `packages/**` (e.g. renaming/removing an export, changing a prop type) can break docs examples or playground usage and still pass every PR gate. The breakage surfaces only post-merge as a failed deploy.

## Location

- `package.json:27` `typecheck` and `package.json:9` `build` are `--filter=./packages/*` only; the always-run PR `build` job (`.github/workflows/pr-checks.yml:63-70`) builds packages only.
- `pr-checks.yml:81-98` `dorny/paths-filter` gates `docs-check` / `playground-check` on `apps/docs/**` / `apps/playground/**` — so a `packages/**`-only PR skips both.
- Vitest is colocated in `packages/0/src` (examples aren't tested). `docs-deploy.yml` / `playground-deploy.yml` run only on push to master (post-merge).

## Impact

CI/DX hygiene, not runtime/security. Coolify keeps serving the last good image, so a failed deploy stalls the update rather than taking the site offline; cost is red CI on master + a fix-forward. Note: `vite-ssg` build does **not** type-check, so type-level example breakage is caught nowhere — the path-filter fix below catches removed/renamed exports (the common case).

## Suggested fix

Add `- 'packages/**'` to **both** the `docs` and `playground` filter lists in `pr-checks.yml` (~`:95-98`). This reuses the existing `docs-check` / `playground-check` jobs (which run `build:docs` / `build:play`), only paying the build cost when packages or the respective app actually change.

sridhar-3009:
## Summary

- Adds `- 'packages/**'` to both the `docs` and `playground` filter lists in the `dorny/paths-filter` step of the `changes` job in `pr-checks.yml`

**Before:** PRs that only touch `packages/*` produce `docs: false` and `playground: false` from the path filter, so neither the docs build nor the playground build runs in CI. Library changes that break app builds go undetected until after merge.

**After:** Any change under `packages/**` also triggers both app builds, ensuring library PRs are validated end-to-end.

Closes #401

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -109,8 +109,10 @@ jobs:
           filters: |
             docs:
               - 'apps/docs/**'
+              - 'packages/**'
             playground:
               - 'apps/playground/**'
+              - 'packages/**'
 
   docs-check:
     needs: changes


johnleider:
Thanks @sridhar-3009 — and appreciate you working the v1.0.0 backlog! This overlaps #428, already open for the same issue (#401) and the version we're landing. Closing as a duplicate to consolidate on #428. Sorry for the parallel effort, and thanks for the contribution. 🙏

PR #464, opened 2026-07-02
feat(Popover): dev-mode warning when display utility overrides native popover hide

### LINKED ISSUE #448: Popover: warn (dev) when a display utility is set on Popover.Content
## Summary

`Popover.Content` renders a native `[popover]` element. The UA stylesheet hides a closed popover via `[popover]:not(:popover-open) { display: none }`. If a consumer sets a `display` utility (`flex` / `grid` / `block`) **directly on `Popover.Content`**, it can override that closed-state hide, so the popover content renders **open on load**.

This is an easy footgun to hit and bit a real example: the Overflow "responsive toolbar" (#419) had `flex flex-col` on `Popover.Content` and rendered the `+N more` menu open — fixed in #444 by moving the display/layout classes to an inner wrapper `<div>`.

## Proposal

A **dev-only** guard/warning (no prod cost) when author CSS sets `display` on the popover content element — steering consumers to wrap their flex/grid layout in a child instead of styling `Popover.Content` directly. Sketch:

- After mount (dev / `__DEV__` only), read the content element's computed `display`; if it's been forced to something other than the popover default while closed, `useLogger().warn(...)` with a pointer to wrap content in a child.
- Headless contract: this should **warn, not override** — v0 must not dictate consumer styling. Dev-only, zero production overhead.

`PopoverContent` already hardcodes `margin: unset` to sidestep native-popover styling pitfalls, so there's precedent for the component accounting for these.

Alternatively (or additionally), document the caveat prominently on the Popover docs page.

Surfaced as a follow-up to #419 / #444.

sridhar-3009:
## Summary

Closes #448.

When a layout utility class (`flex`, `grid`, `inline-flex`, etc.) is placed directly on `<Popover.Content>`, it can conflict with the UA `[popover]:not(:popover-open) { display: none }` rule that hides a closed popover. This was the root cause of #419; the CSS-layer fix landed in #419/398, but there was no developer feedback to prevent the pattern from reappearing.

### What this PR adds

**`PopoverContent.vue`** — `onMounted` guard (tree-shaken to nothing in production via `typeof __DEV__ !== 'undefined' && __DEV__`):

1. Clones the content element and removes its `[popover]` attribute, so neither the UA closed-state rule nor the component's own `[popover]:not(:popover-open){display:none!important}` override applies to the clone.
2. Appends the clone off-screen and reads `getComputedStyle(clone).display`.
3. If the computed value is anything other than `none` or `block` (i.e. the default div display), a `useLogger('PopoverContent').warn(...)` is emitted, naming the detected value and pointing to the fix: wrap layout classes in a child element.
4. Removes the clone immediately — zero DOM leak.

The guard is also SSR-safe via the `IN_BROWSER` early return.

### Tests

Two new cases in `index.test.ts` under `content > dev warning`:

- **warns** when `getComputedStyle` returns `'flex'` for the off-screen probe (simulated via `vi.spyOn`).
- **does not warn** when `getComputedStyle` returns the happy-dom default `''` (no utility class present).

### Before / After

```html
<!-- triggers warning in dev -->
<Popover.Content class="flex flex-col gap-4">…</Popover.Content>

<!-- correct: layout on the inner wrapper -->
<Popover.Content>
  <div class="flex flex-col gap-4">…</div>
</Popover.Content>
```

### DIFF
--- a/packages/0/src/components/Popover/PopoverContent.vue
+++ b/packages/0/src/components/Popover/PopoverContent.vue
@@ -46,8 +46,14 @@
   // Context
   import { usePopoverContext } from './PopoverRoot.vue'
 
+  // Composables
+  import { useLogger } from '#v0/composables/useLogger'
+
+  // Globals
+  import { IN_BROWSER } from '#v0/constants/globals'
+
   // Utilities
-  import { toRef, useTemplateRef } from 'vue'
+  import { onMounted, toRef, useTemplateRef } from 'vue'
 
   defineOptions({ name: 'PopoverContent' })
 
@@ -85,6 +91,35 @@
   })
 
   context.attach(() => ref.value?.element)
+
+  if (typeof __DEV__ !== 'undefined' && __DEV__) {
+    const logger = useLogger('PopoverContent')
+
+    onMounted(() => {
+      if (!IN_BROWSER) return
+      const el = ref.value?.element
+      if (!el) return
+
+      // Clone the element without [popover] so neither the UA closed-state rule
+      // nor our [popover]:not(:popover-open){display:none!important} override
+      // applies. A non-block computed display on the clone signals that an author
+      // utility class (flex, grid, …) has been placed directly on the element.
+      const probe = el.cloneNode() as HTMLElement
+      probe.removeAttribute('popover')
+      probe.style.position = 'fixed'
+      probe.style.visibility = 'hidden'
+      probe.style.pointerEvents = 'none'
+      probe.style.top = '-9999px'
+      document.body.append(probe)
+      const display = window.getComputedStyle(probe).display
+      probe.remove()
+
+      if (display && display !== 'none' && display !== 'block') {
+        const msg = `[Popover.Content] A display utility class (detected: "${display}") is set directly on the content element. This conflicts with the native [popover] closed-state hide. Move layout classes (flex, grid, etc.) to a wrapper element inside the slot instead.`
+        logger.warn(msg)
+      }
+    })
+  }
 
   function onBeforeToggle (e: ToggleEvent) {
     emit('beforetoggle', e)
--- a/.changeset/popover-display-utility-dev-warning-448.md
+++ b/.changeset/popover-display-utility-dev-warning-448.md
@@ -1,0 +1,7 @@
+---
+"@vuetify/v0": patch
+---
+
+feat(Popover): dev-mode warning when a display utility overrides native popover hide
+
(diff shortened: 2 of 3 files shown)


johnleider:
I wonder if this would be better served as a warning to the user. It's definitely a good conversation to have regarding when/where/why we would add behavior like this to a component. @J-Sek @AndreyYolkin @Haviles04

packages/0/src/components/Popover/PopoverContent.vue
```diff
@@ -86,6 +92,35 @@
...
+      probe.style.position = 'fixed'
+      probe.style.visibility = 'hidden'
+      probe.style.pointerEvents = 'none'
+      probe.style.top = '-9999px'
+      document.body.append(probe)
+      const display = window.getComputedStyle(probe).display
+      probe.remove()
+
+      if (display && display !== 'none' && display !== 'block') {
+        const msg = `[Popover.Content] A display utility class (detected: "${display}") is set directly on the content element. This conflicts with the native [popover] closed-state hide. Move layout classes (flex, grid, etc.) to a wrapper element inside the slot instead.`
```
J-Sek, line 118: code probes computed style, not class, so "_A display utility class ..._"  is not precise and might be misleading.

J-Sek:
It is hard to take a stance - scenario suggests, the menu is being rendered, so why not just assume it is likely to be visible, making console warning redundant. It only potentially serves people that ship straight to Prod without QA and skipping manual verification locally. That's also a bit weak.. they probably skip opening browser and running affected web page altogether, so console logs won't change much.

I would prefer ESLint rule, but can't help by notice testing against computed style covers more ground. The only issue is potential performance overhead.

Vue DevTools already hit hard, so maybe it does not matter, but we could make sure by having 20x500 table rows here each table has some kind of dropdown for context menu or value selector.

We can also make it lighter by setting empty inner body `probe.innerHTML = ''`

My vote would be to skip this case. Simply rely on class making the popover visible.

packages/0/src/components/Popover/PopoverContent.vue
```diff
@@ -86,6 +92,35 @@
...
+  if (typeof __DEV__ !== 'undefined' && __DEV__) {
+    const logger = useLogger('PopoverContent')
+
+    onMounted(() => {
+      if (!IN_BROWSER) return
+      const el = ref.value?.element
+      if (!el) return
+
+      // Clone the element without [popover] so neither the UA closed-state rule
+      // nor our [popover]:not(:popover-open){display:none!important} override
```
J-Sek, line 104: I don't think v0 ships these styles + if UA makes the display hidden, then it wins over class and there is no issue to warn about - i.e. cloning is adding friction instead of solving real problem.

Haviles04:
I agree with Jacek that an ESLINT rule is preferable, this feels heavy just to say "hey maybe don't do that"

johnleider:
Thank you for the Feature Request and interest in improving Vuetify. After careful consideration, the team has decided that this is not functionality that we are looking to implement at this time.

If you have any questions, please reach out to us in our [Discord community](https://community.vuetifyjs.com).

PR #539, opened 2026-07-07
docs(benchmarks): tier on per-unit cost with worst-of roll-up

johnleider:
## What

Replaces hz-threshold benchmark tiering with a cost model, and fixes the docs badge roll-up that reported every feature as its *fastest* bench's tier.

### Tiering (`scripts/lib/benchmarks.ts`)

Raw ops/s measures workload size, not code speed. Workload (item count × repeat factor) is parsed from the bench name.

**Collection benches (items > 1)** — worst of two axes:

| Axis | blazing | fast | good | slow |
|---|---|---|---|---|
| Per-item cost | <1μs | <10μs | <100μs | ≥100μs |
| Single-op latency (mean ÷ repeats) | ≤1 frame (16.7ms) | ≤2 frames | ≤100ms | >100ms |

The latency axis keeps a 160ms 10k-row sort from badging fast on honest per-item cost alone.

**One-shot benches (items = 1: constructors, single utility calls)** tier on call latency alone — per-item budgets at items=1 are a category error that graded a 22μs constructor C-tier:

| blazing | fast | good | slow |
|---|---|---|---|
| <10μs | <100μs | ≤1 frame | beyond |

`detectComplexity` (name-guessed O(1)/O(n)/O(n²)) is removed.

### Roll-up

- Group tier = **worst** bench in the group (was: score average)
- Feature tier (new `_tier` field) = worst group
- Docs badge reads `_tier` (was: `_fastest.tier` — all 12 features badged **blazing**)
- Missing metrics render an explicit **unmeasured** state (was: silent `?? 'good'`)

### Data

Committed metrics artifacts (current + 7 per-version history snapshots) retiered in place from stored hz/mean with a scale guard (hz×mean≈1000); coverage untouched. CI metrics-regen reproduces identical tiers via the shared lib on the next release.

## Resulting feature badges

| Feature | Before | After |
|---|---|---|
| createRegistry, createVirtual | blazing | blazing |
| createDataTable, createModel, createSelection, createSortable, createTokens, useDate, useProxyRegistry | blazing | fast |
| createFilter | blazing | **good** (updating a filter query on 10k items costs 48ms — user-perceptible while typing) |
| createNested | blazing | **good** (onboarding 10k nodes costs 70ms) |
| createDataGrid | blazing | **slow** (single 10k-row sort/search ops cost 100–168ms per call) |

Bench-level distribution: 221 blazing / 120 fast / 20 good / 7 slow (was 165/60/102/41 with the false-positive skew documented in the rc.6 investigation). The remaining good/slow entries are genuine optimization targets, not workload-size artifacts.

## Notes for review

- `createRegistry`'s reactive `values()/keys()/entries()` benches now tier **blazing** (0.3μs per element-visit, ~3.3ms per call at 10k). The old model tagged them slow. Per-item cost says the per-element price is fine; the O(n)-per-call design is the documented #209 cache-skip tradeoff. If we want the badge to punish the algorithmic choice rather than the per-element cost, that's a different (third) axis — flagging explicitly rather than hiding it.
- Workload parsing is name-convention-based (`(10,000 items)`, `100 times`, bare numbers ≥1000). Unparsed quantities default to 1, which over-estimates cost — errs strict.
- Follow-up (separate PR): noise-aware regression gate comparing releases via the fixed-apparatus history.


johnleider:
Note: commit fbf3ecc28 (the createRegistry reactive-read fix) was built and verified on this branch, then extracted to #540 as the standalone packages/0 change. Rebase this branch once #540 merges — the duplicate commit will drop out.

PR #546, opened 2026-07-08
fix(Combobox): route ComboboxEmpty default slot text through useLocale

### LINKED ISSUE #528: fix(Combobox): route ComboboxEmpty default through useLocale
`ComboboxEmpty.vue:58` renders a hardcoded English default:

```vue
<slot v-bind="slotProps">No results</slot>
```

This is a user-facing string with no `useLocale` path, which PHILOSOPHY §5.5 ("Locale-first strings") forbids. Every other component resolves such strings via the canonical idiom, e.g. `DialogClose.vue:68` → `locale.ti('Dialog.close') ?? 'Close'`.

### Fix
- Add a `Combobox.noResults` key to `packages/0/src/locale/messages/en/index.ts`.
- Change the slot default to `{{ locale.ti('Combobox.noResults') ?? 'No results' }}` (resolve `locale` in setup like the sibling components).

Consumers still override via the default slot; this only localizes the fallback.

### Notes
Surfaced during a PHILOSOPHY.md consistency audit. The doc rule (§5.5) is correct and stays as-is — this brings the source into compliance rather than weakening the rule. `ProgressRoot`'s `aria-valuetext` `${Math.round(pct)}%` was also flagged but is number formatting, not a translatable phrase — out of scope here.

sridhar-3009:
## Summary

Closes #528.

`ComboboxEmpty.vue:58` rendered a hardcoded English `"No results"` fallback with no `useLocale` path, violating PHILOSOPHY §5.5 (locale-first strings). Every other component resolves such strings via `locale.ti('Component.key') ?? 'fallback'`.

### Changes

- **`packages/0/src/locale/messages/en/index.ts`** — Add `Combobox.noResults: 'No results'` in alphabetical order (between `Button` and `Carousel`).
- **`ComboboxEmpty.vue`** — Import `useLocale`, call it in setup, replace the slot default text with `{{ locale.ti('Combobox.noResults') ?? 'No results' }}`.

Consumers who supply a `default` slot override are completely unaffected — the change only touches the fallback rendered when no slot content is provided.

### Tests

Two new cases in `index.test.ts` under `filtering`:

- **No plugin installed** — types a no-match query, asserts `Combobox.Empty` renders `"No results"` (the hardcoded `??` fallback).
- **Plugin installed** — installs a `fr` locale with `Combobox.noResults: 'Aucun résultat'`, types a no-match query, asserts the component renders the translated string.

### DIFF
--- a/packages/0/src/components/Combobox/index.test.ts
+++ b/packages/0/src/components/Combobox/index.test.ts
@@ -521,7 +521,7 @@ describe('combobox', () => {
       expect(empty.exists()).toBe(true)
     })
 
-    it('should fall back to hardcoded "No results" when no locale plugin is installed', async () => {
+    it('should fall back to the inline default "No results" when no locale plugin is configured', async () => {
       const wrapper = mount(
         defineComponent({
           render: () =>
@@ -551,7 +551,7 @@ describe('combobox', () => {
       expect(empty.text()).toBe('No results')
     })
 
-    it('should use locale-translated string when locale plugin is installed', async () => {
+    it('should use the translated string when locale plugin is configured', async () => {
       const plugin = createLocalePlugin({
         default: 'fr',
         messages: {
--- a/packages/0/src/components/Combobox/index.test.ts
+++ b/packages/0/src/components/Combobox/index.test.ts
@@ -3,6 +3,9 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 // Adapters
 import { ClientComboboxAdapter } from '#v0/composables/createCombobox/adapters/client'
 
+// Composables
+import { createLocalePlugin } from '#v0/composables'
+
 import { Combobox } from './index'
 
 // Utilities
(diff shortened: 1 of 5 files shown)


packages/0/src/locale/messages/en/index.ts
```diff
@@ -31,6 +31,9 @@ export default {
   Button: {
     label: 'Button',
   },
+  Combobox: {
```
johnleider: **[suggestion]** The `Combobox` entry is inserted between `Button` and `Carousel`. The file (and the parallel list in its sibling test) maintains top-level namespace keys in alphabetical order; `Carousel` (Ca) must precede `Combobox` (Co).

**Suggestion:** Move the `Combobox: { noResults: 'No results' }` block after the entire `Carousel` block (before `Dialog`) to match the "(alphabetical)" claim in the changeset/PR body and preserve existing ordering.

packages/0/src/components/Combobox/index.test.ts
```diff
@@ -518,6 +521,77 @@ describe('combobox', () => {
       expect(empty.exists()).toBe(true)
     })
 
+    it('should fall back to hardcoded "No results" when no locale plugin is installed', async () => {
```
johnleider: **[nit]** The first new test is named `should fall back to hardcoded "No results" when no locale plugin is installed`. All equivalent fallback tests in the suite (Pagination, Carousel, Rating, Snackbar, Dialog, AlertDialog, ...) consistently use the wording "fall back to the inline default".

**Suggestion:** Rename to `should fall back to the inline default "No results" when no locale plugin is configured` (or similar) for consistency with precedent.

johnleider (approved):

Issue #339, opened 2026-06-12
fix(useRtl,useTheme): adopt the leak-safe adapter lifecycle shipped in useReducedMotion


johnleider:
## Problem

`useRtl` and `useTheme` adapters register their reactive watches with `onScopeDispose(stop, true)` from the plugin install path — which runs outside any effect scope, so the registration silently no-ops. Dir/CSS watches and unhead entries are **never torn down on `app.unmount`**: matchMedia listeners and watchers leak under repeated app create/destroy (tests, SSR-per-request, micro-frontends).

`useReducedMotion` (#278) ships the corrected lifecycle; this issue is the family sweep.

## What to copy from useReducedMotion

1. **Leak-safe teardown** — stash watch-stops/unhead-entry teardown on `adapter.dispose`; the plugin registers `app.onUnmount(() => adapter.dispose?.())` (late-bound, unconditional). Expose `dispose()` on the public context so non-plugin paths (`createXContext`, standalone factories) have a reachable handle.
2. **Typed unhead seam** — `app._context.provides.usehead` is an unchecked `any` in `useRtl/adapters/v0.ts`, `useTheme/adapters/v0.ts`, and `useTheme/adapters/unhead.ts`. Copy the structural `Head`/`HeadEntry` duck-types from `useReducedMotion/adapters/v0.ts`.
3. **Reactive SSR head entries** — patch via `entry.patch?.()` where state can change during server render (useTheme's unhead adapter currently pushes once, statically).
4. **Validated restore** — replace the blind casts (`saved as ID` in useTheme/useLocale, `saved as boolean` in useRtl) with literal/type-guard validation per the new `composables.md` "Persist/restore validation" rule.
5. **Decide the `provides.head` fallback** — all four resolution sites also probe the plain `head` key; no known framework provides it (@unhead/vue v1/v2 and @vueuse/head all use `usehead`). Keep or drop consistently across the family.

`.claude/rules/implementation.md` "Adapter Pattern" documents the target lifecycle; `useReducedMotion` is the reference implementation. Add the regression test shape from its suite (custom adapter, assert `dispose` called once on `app.unmount`).

PR #605, opened 2026-07-15
chore(ci): add bundle-size gate for @vuetify/v0 key exports (#543 WS2)

sridhar-3009:
## Summary

Adds bundle-size CI gate to catch accidental regressions before they land — part of #543 WS2 (RC → 1.0 hardening).

- Adds `.size-limit.json` at repo root with per-export budgets for 7 entry points: `createRegistry`, `createSelection`, `useStack`, `useTheme`, `usePopover`, `createDataTable`, and a combined barrel ceiling
- Adds `pnpm size` script (wraps `size-limit`) and a `size` CI job to `pr-checks.yml`
- The `size` job fires only on `packages/**` path changes (new `packages` output on the existing `changes` job) and reuses the `packages-dist` artifact already uploaded by the `build` job — no double build
- `size-limit` + `@size-limit/preset-small-lib` added to catalog and root devDeps; preset added to knip `ignoreDependencies` (loaded implicitly by size-limit)

## Budget methodology

Budgets seeded from `pnpm build:0` dist output (`1.0.0-rc.7`), measured with esbuild (`vue` external, minified + gzipped), rounded to nearest 0.5 kB with a ~5% buffer:

| Export | Measured (gzip) | Limit |
|---|---|---|
| `createRegistry` | 4.02 kB | 4.5 kB |
| `createSelection` | 4.92 kB | 5.5 kB |
| `useStack` | 5.88 kB | 6.5 kB |
| `useTheme` | 7.73 kB | 8.5 kB |
| `usePopover` | 6.80 kB | 7.5 kB |
| `createDataTable` | 10.50 kB | 11.5 kB |
| barrel (6 named exports from root) | 15.37 kB | 17 kB |

All measurements exclude `vue` (peer dependency). The gate fails when any single import grows more than ~10% above the current baseline.

## Test plan

- [x] CI `size` job passes on this PR (no packages changed, so `size` is skipped — verify `changes.outputs.packages == 'false'`)
- [x] Verify a PR that changes `packages/0/src/**` triggers the `size` job
- [x] Manually verify: intentionally bloating a composable by +10 kB should red the gate

Closes #543 (WS2)

Issue #406, opened 2026-06-29
[Bug Report] useTheme: cspNonce dropped on SSR head path; unhead adapter ignores it


johnleider:
## Summary

`V0StyleSheetThemeAdapter` accepts a `cspNonce` option but never applies it. On the SSR head path it pushes a `<style>` without the nonce, and `V0UnheadThemeAdapter` doesn't accept a nonce at all. Strict-CSP apps get their SSR theme styles blocked (FOUC), and the stored option gives false assurance.

## Location

- `packages/0/src/composables/useTheme/adapters/v0.ts:31,41` declare `cspNonce`; `:46` stores `this.cspNonce = options.cspNonce` — but it is never read again.
- SSR branch (`:94-98`) resolves `head` and calls `head.push({...})` without threading the nonce into the style entry.
- Client path (`:122-129`, `adoptedStyleSheets`) correctly needs no nonce.
- `adapters/unhead.ts` does not accept `cspNonce`.

## Impact

Low (medium for strict-CSP SSR). With `style-src 'nonce-…'`, the SSR `<style>` is blocked → FOUC until client hydration swaps to `adoptedStyleSheets`. Accepted-but-unused config is a wart: honor it or remove it.

## Suggested fix

Thread `cspNonce` into the SSR `head.push` style entry (as a `nonce` attribute), and accept/forward a nonce in the unhead adapter. Add a test asserting the SSR `<style>` carries the nonce.

PR #606, opened 2026-07-15
fix(useRtl,useTheme): add reactive SSR head entry updates via entry.patch (#543 WS3)

### LINKED ISSUE #543: [Chore] RC → 1.0 hardening: freeze-confidence for the stable set
Tracking issue for RC → 1.0 hardening. Derived from a 3-week commit audit (2026-06-16 → 07-07): the sprint was docs/release-heavy (88 docs, 36 chore, 24 fix, **only 2 perf, 2 test, 1 refactor**). Source is clean — this is about **freeze-confidence for the stable set**, not refactoring.

Ordered by leverage. WS1/WS2/WS3 are file-independent and can run in parallel worktrees. **Out of scope:** speculative refactors and perf micro-tuning (5 debt markers + 3 ts-suppressions total across the package — nothing to chase).

---

### WS1 — Benchmark the unbenched stable selection primitives · ~half day · do first
`createGroup` / `createSingle` / `createStep` are promoted `stable` (frozen API, break = major) but have **zero benchmarks**. The whole selection tree (Tabs, Treeview, Carousel, forms) sits on them — no baseline means a post-1.0 patch can silently regress them.

- [x] `packages/0/src/composables/createGroup/index.bench.ts`
- [x] `packages/0/src/composables/createSingle/index.bench.ts`
- [x] `packages/0/src/composables/createStep/index.bench.ts`
- [x] Mirror `createSelection/index.bench.ts`; read `.claude/rules/benchmarks.md` (fixture isolation mandatory); datasets 1k + 10k
- [x] Confirm hot paths from each `index.ts` before finalizing categories (e.g. `createSingle` deselect-previous)
- [x] `pnpm metrics` to regen tier badges
- **Accept:** `pnpm test:bench` green, `benchmarks.json` updated, badges render, no source changes

### WS2 — Bundle-size gate in CI · ~half day · do second
#347/#443 changed `sideEffects`, web-types, and browser-entry declarations — the exact tree-shaking levers — with **no size check anywhere**. Highest silent risk for a headless "import only what you use" lib.

- [x] Add `size-limit` + `@size-limit/preset-small-lib` (pnpm devDeps)
- [x] Per-export budgets: `createRegistry`, `createSelection`, `useTheme`, `useStack`, `usePopover`, `createDataTable` + a bare-barrel ceiling
- [x] Seed budgets from current `build:0` output, +~5% for noise
- [x] CI job on `packages/**` path filter (mirror #428)
- **Accept:** `pnpm size` passes; fails on intentional +10% bloat; single named import proven not to drag the barrel

### WS3 — Root-cause audit: `useStack` + `useTheme` · ~half day · highest bug-prevention
3 symptom-fixes each in 3 weeks = one structural flaw underneath. Fix #4/#5 will arrive from users post-launch unless audited now.

- [x] `useStack`: map register→activate→block/scrim→dispose; does one lifecycle invariant subsume #418/#442/#354?
- [x] `useTheme`: adapter dispose + SSR head path — is mount/dispose symmetric? Cross-ref adapter dispose-leak (#338/#339)
- [x] Grep siblings `useRtl`/`useLocale` (same adapter pattern) for the dispose leak
- [x] Write findings **before** refactoring
- **Accept:** root-cause note; either a targeted fix PR or an evidenced "independent fixes, no shared cause"

### WS4 — ARIA/keyboard sweep across compound components · ~1 day · can slip to day 2
§1 claims WAI-ARIA compliance; Button/Treeview/`ti()` were point fixes, never a systematic pass.

- [x] Audit 40 components against their APG pattern (roles, aria wiring, roving tabindex, focus return, key handling)
- [x] Prioritize disclosure + forms families (Dialog, Popover, Tabs, Treeview, Combobox, Select, Checkbox, Radio, Switch)
- [x] Land cheap wins inline; defer deep ones to individual issues on this milestone
- **Accept:** completed audit matrix (component × APG requirement × pass/fail)

### WS5 — SSR render smoke test · ~half day · nice-to-have pre-1.0
SSR is spot-fixed (#442, #441), never verified. Vapor parity harness (#420) is the template.

- [x] Isolated SSR harness rendering each browser-global-touching composable; assert no throw + no hydration mismatch
- [x] Harvest the prototype on branch `test/ssr-hydration-proto`
- **Accept:** `pnpm test:ssr` green

---

Run order: **AM** WS1 + WS2 (parallel sessions) · **PM** WS3 · **Day 2** WS4, WS5.

sridhar-3009:
## WS3 Audit Findings

Part of #543 — RC → 1.0 hardening. Documents lifecycle audit results and fixes the one remaining gap found.

### useStack lifecycle (#354 / #418 / #442) — ✅ no shared cause

All three fixes are present and independent:

| Issue | Root cause | Fix | Status |
|---|---|---|---|
| #354 | `isActive`/`top` counted non-modal (`scrim: false`) tickets | `selectedScrims()` filters to scrim-bearing subset | ✅ fixed |
| #418 | `blocking`/`scrim` snapshotted at register time (non-reactive) | Normalize to `toRef(() => toValue(input.blocking) ?? false)` | ✅ fixed |
| #442 | Module-scoped `fallbackStack` leaked across SSR requests | Return ephemeral `createStack()` per call under `!IN_BROWSER` | ✅ fixed |

No shared invariant — three independent symptom/fix pairs. Fix #4 is unlikely from the same source.

### useTheme + useRtl lifecycle (#338 / #339) — ✅ mostly fixed, one residual gap

| Item | Status |
|---|---|
| #338 lazy plugin context | ✅ fixed — context created inside `provide` at install time |
| #339 `app.onUnmount(() => adapter.dispose?.())` | ✅ fixed in both `useTheme` and `useRtl` |
| #339 `restore` validation (blind casts removed) | ✅ fixed — `typeof saved === 'boolean'`/`string\|number` guards |
| **#339 reactive SSR head entries via `entry.patch`** | ❌ **not yet fixed** — this PR fixes it |

The `useReducedMotion` reference sets up a watch in the SSR path to call `entry.patch?.()` when state changes. Both `useTheme` (v0 + unhead adapters) and `useRtl` pushed once statically, leaving the head entry stale during async server renders.

### useLocale — ✅ no adapter lifecycle

`useLocale` is a stateless translation utility. No `adapter.setup()`/`adapter.dispose()` pattern, no live resources to audit. `restore` already uses type-guard validation.

## Fix (this PR)

Three files, same pattern as `useReducedMotion`:

- **`useRtl/adapters/v0.ts`**: add `patch?` to `HeadEntry`; set up watch on `context.isRtl` in SSR path; composite dispose stops the watch AND calls `entry.dispose?.()`.
- **`useTheme/adapters/v0.ts`**: same — `patch?` on `HeadEntry`; watch `[selectedId, colors, isDark]`; composite dispose.
- **`useTheme/adapters/unhead.ts`**: SSR `else` branch had no watch at all; add it so `V0UnheadThemeAdapter` also stays live during async server rendering.

## Test plan

- [x] `pnpm typecheck:0` clean
- [x] Existing `useRtl` + `useTheme` tests pass
- [x] Manually verify: set up unhead + SSR, change theme/dir during render, confirm `data-theme`/`dir` reflects the change in the SSR HTML

Closes #543 (WS3 — root-cause audit complete, residual SSR head-entry gap fixed)


johnleider:
This branch was cut from a stale master that included an older copy of #605's commit, which is why the diff showed the WS2 files and CI never ran. Pushed a merge of the current #605 branch to bring it up to date — the diff collapses to the WS3 fix only once #605 merges, so #605 should land first.

Two things worth review on top of the original fix:

- conflict resolution in `useTheme/adapters/v0.ts`: master added a CSP nonce to the pushed style entry (#406); I kept it and mirrored it into the new `entry.patch` payload so a patch doesn't strip the nonce
- the three SSR dispose tests asserted `adapter.dispose` was `entry.dispose` by identity; updated them to assert the new composite dispose delegates to `entry.dispose` and exists even when the entry has none (the watch still needs cleanup). The patch-on-change behavior itself has no direct test yet.

johnleider (approved):

PR #618, opened 2026-07-15
fix(Avatar,Scrim,Popover,Tooltip,Select,Toggle): restore dropped alt text and complete missing ARIA wiring (#543)

johnleider:
Six small, verified accessibility fixes from the WS4 ARIA audit. All changes are additive attribute wiring; no behavior changes outside ARIA/attrs.

| Fix | Component | Evidence | Change |
|---|---|---|---|
| 1 | `Avatar.Image` | `inheritAttrs: false` + `v-bind="slotProps.attrs"` without `useAttrs()` merge dropped consumer `alt`/`aria-label` despite hardcoded `role="img"` | Mirror `ImageImg`: attrs fall through, explicit `alt` prop |
| 2 | `Scrim` | `getSlotProps` bound only `style`/`onClick` — decorative backdrop exposed to AT | `aria-hidden="true"` on each layer |
| 3 | `Popover.Activator` | Only `popovertarget`/`type` bound; native implicit ARIA mapping is inconsistent | Explicit `aria-expanded` from `context.isOpen`, `aria-controls` -> content id (`target ?? context.id`) |
| 4 | `Tooltip.Content` | Activator handled Escape, content had no keydown — Escape dead with focus inside interactive tooltips | Same Escape-close handler on content |
| 5 | `Select` | Activator lacked disabled ARIA + focusability fallback; listbox unnamed | `aria-disabled` + native `disabled` (mirrors `ComboboxControl`), `tabindex="0"` for non-button `as`, `aria-labelledby` -> activator id on content (mirrors `ComboboxContent`) |
| 6 | `Toggle.Group` | `role="group"` with no naming surface | `label` / `ariaLabelledby` / `ariaDescribedby` props, mirroring `ButtonGroup` exactly |

Refs #543 (WS4)

### DIFF
--- a/.changeset/a11y-overlays-543.md
+++ b/.changeset/a11y-overlays-543.md
@@ -1,0 +1,12 @@
+---
+'@vuetify/v0': patch
+---
+
+fix(Avatar,Scrim,Popover,Tooltip,Select,Toggle): restore dropped alt text and complete missing ARIA wiring (#543)
+
+- `Avatar.Image` now accepts an `alt` prop and passes consumer attributes (`alt`, `aria-label`, ...) through to the rendered element — previously they were silently dropped
+- `Scrim` backdrops are hidden from assistive technology with `aria-hidden="true"`
+- `Popover.Activator` explicitly exposes `aria-expanded` and `aria-controls` instead of relying on inconsistent native `popovertarget` mapping
+- `Tooltip.Content` closes on Escape when focus is inside interactive tooltip content
+- `Select.Activator` reflects the disabled state (`aria-disabled` + native `disabled`) and stays keyboard-focusable when rendered as a non-button element; `Select.Content` names its listbox via `aria-labelledby`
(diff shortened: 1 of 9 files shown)


johnleider:
Second-pass review: clean, no blockers. Two notes for the record:

- `Scrim`'s new `aria-hidden` is on the slot-bearing root — correct for the pure-backdrop case, but focusable content in the default slot would violate WCAG 4.1.2. Worth a one-line docs note that the slot isn't for interactive content.
- `SelectActivator`'s tabindex fallback keeps a non-button activator focusable while disabled (native `disabled` no-ops on non-buttons). ARIA-legal (focusable + aria-disabled), and `open()` is guarded, so intentional.

johnleider:
Docs note added in f0b2e96a0 — `scrim.md` now calls out that the aria-hidden slot isn't for interactive content.

PR #619, opened 2026-07-15
fix(Switch,Form,Slider): correct ARIA states for mixed, native validation, and grouped form controls (#543)

johnleider:
Lands the verified fixes from the WS4 ARIA audit, forms family. Each finding was re-verified against source before changing anything; one audit finding did not hold and was skipped (table below).

| # | Finding | File(s) | Outcome |
|---|---------|---------|---------|
| 1 | `aria-invalid`/`aria-errormessage` desync with manual `errorMessages` | `packages/0/src/components/Input/InputControl.vue`, `NumberField/NumberFieldControl.vue` | **Skipped — not reproducible.** `createInput`'s `isValid` already returns `false` for manual-only errors (`composables/createInput/index.ts` ~162: `if (errors.value.length > 0 && validation.errors.value.length === 0) return false`), and `createValidation` sets `errors`/`isValid` atomically, so `aria-errormessage` presence always implies `aria-invalid="true"`. |
| 2 | Spec-invalid `aria-checked="mixed"` on `role=switch` | `Switch/SwitchRoot.vue`, `Switch/SwitchSelectAll.vue` | Clamped to `false` while indeterminate (ARIA 1.2 forbids `mixed` on switch); `data-state="indeterminate"` unchanged. Existing browser tests updated to assert the clamp. |
| 3 | Decorative thumb/track exposed to AT | `Switch/SwitchThumb.vue`, `Switch/SwitchTrack.vue` | Added `aria-hidden="true"` to bound attrs. |
| 4 | Missing `novalidate` — native constraint UI preempts async validation | `Form/Form.vue` | Added `novalidate: true` to form attrs. |
| 5 | Multi-thumb slider root has no group semantics | `Slider/SliderRoot.vue` | Added `role="group"` + optional `label`/`ariaLabelledby` props, mirroring the CheckboxGroup/ButtonGroup pattern. |

Refs #543 (WS4)

### DIFF
--- a/.changeset/a11y-forms-543.md
+++ b/.changeset/a11y-forms-543.md
@@ -6,5 +6,5 @@ fix(Switch,Form,Slider): correct ARIA states for mixed, native validation, and g
 
 - `Switch.Root` and `Switch.SelectAll` no longer emit the spec-invalid `aria-checked="mixed"`; the value is clamped to `false` while indeterminate, so screen readers announce a valid switch state. Style indeterminate switches with `data-state="indeterminate"` as before.
 - `Switch.Thumb` and `Switch.Track` are now marked `aria-hidden="true"`, hiding the decorative visuals from assistive technology.
-- `Form.Root` now renders `novalidate`, so the browser's native constraint popups no longer block submit before v0's async validation runs.
+- `Form.Root` now renders `novalidate` by default, so the browser's native constraint popups no longer block submit before v0's async validation runs. Opt back into native constraint validation with `:novalidate="false"`.
 - `Slider.Root` now exposes `role="group"` plus optional `label` / `ariaLabelledby` props, giving multi-thumb sliders an accessible group name.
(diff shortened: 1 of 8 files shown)


johnleider:
Second-pass review outcome:

- **Applied**: `novalidate` is now a prop defaulting `true` instead of a hardcode — `:novalidate="false"` restores native constraint validation (the merge order previously made it un-overridable). Changeset updated.
- **Applied**: Switch clamp simplified to `!isMixed.value && isChecked.value`.
- **Open note**: Slider's new `label`/`ariaLabelledby` props + `role="group"` and the Form `novalidate` attr ship without new assertions (per repo tests-only-when-asked convention). Say the word if you want coverage added here.

Switch clamp verified spec-correct: `mixed` is invalid on `role=switch`; `data-state=indeterminate` styling contract untouched, and the updated browser assertions keep the indeterminate check.

PR #622, opened 2026-07-15
fix dangling aria idrefs

### LINKED ISSUE #608: [Bug] a11y: unconditional aria-labelledby/describedby emit dangling IDREFs (Dialog, AlertDialog, Progress)

sridhar-3009:
## Summary

Fixes #608.

Dialog, AlertDialog, and Progress now only emit ARIA IDREF attributes after the referenced optional subcomponent is mounted. That prevents aria-labelledby and aria-describedby from pointing at elements that do not exist when title, description, or label components are omitted.

## Changes

- Track mounted Dialog and AlertDialog title/description components in root context.
- Track mounted Progress label components in root context.
- Omit aria-labelledby / aria-describedby when the referenced target is missing.
- Add regression coverage for both present and missing target branches.

## Validation

- pnpm vitest run packages/0/src/components/Dialog/index.browser.test.ts packages/0/src/components/AlertDialog/index.browser.test.ts packages/0/src/components/Progress/index.test.ts
- pnpm --filter=./packages/0 typecheck

Note: local Node is v25.6.1, so pnpm reports the repo engine warning for Node >=26, but the focused checks passed.

### DIFF
--- a/packages/0/src/components/AlertDialog/AlertDialogContent.vue
+++ b/packages/0/src/components/AlertDialog/AlertDialogContent.vue
@@ -48,8 +48,8 @@
       'id': string
       'role': 'alertdialog'
       'aria-modal': 'true'
-      'aria-labelledby': string
-      'aria-describedby': string
+      'aria-labelledby': string | undefined
+      'aria-describedby': string | undefined
       'style': { zIndex: number }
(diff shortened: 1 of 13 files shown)


sridhar-3009:
Closing — created via Codex, not ready.

PR #623, opened 2026-07-15
add default splitter handle label

### LINKED ISSUE #617: [Feature] Splitter: default accessible name for the handle via locale
From the #543 WS4 audit. SplitterHandle.vue:271 emits `aria-label` only when the consumer passes `label` — an unnamed `role=separator` otherwise. Add a localized default (new locale message key + fallback), mirroring how ButtonRoot auto-labels icon-only buttons (ButtonRoot.vue:217). Optionally add `aria-valuetext` for percentage announcements. Everything else in the window-splitter pattern is complete (full keyboard incl. PageUp/Down/Home/End/Enter-collapse, SplitterHandle.vue:177-236).

sridhar-3009:
## Summary

Fixes #617.

Splitter handles now get a default accessible name when consumers do not pass the `label` prop. The explicit prop still wins, while the fallback uses the locale catalog with a built-in English fallback.

## Changes

- Add `Splitter.handle` to the English locale messages.
- Use `locale.ti('Splitter.handle')` for the default Splitter handle `aria-label`.
- Keep custom `label` prop behavior unchanged.
- Add regression coverage for default and custom labels.

## Validation

- pnpm vitest run packages/0/src/components/Splitter/index.browser.test.ts packages/0/src/locale/messages/en/index.test.ts
- pnpm --filter=./packages/0 typecheck

Note: local Node is v25.6.1, so pnpm reports the repo engine warning for Node >=26, but the focused checks passed.

### DIFF
--- a/packages/0/src/components/Splitter/SplitterHandle.vue
+++ b/packages/0/src/components/Splitter/SplitterHandle.vue
@@ -17,6 +17,7 @@
 
   // Composables
   import { useDocumentEventListener } from '#v0/composables/useEventListener'
+  import { useLocale } from '#v0/composables/useLocale'
   import { useRaf } from '#v0/composables/useRaf'
   import { useToggleScope } from '#v0/composables/useToggleScope'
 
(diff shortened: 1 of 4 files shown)


sridhar-3009:
Closing — created via Codex, not ready.

PR #631, opened 2026-07-17
fix(Dialog,AlertDialog): omit aria-labelledby/describedby when title/description absent

### LINKED ISSUE #608: [Bug] a11y: unconditional aria-labelledby/describedby emit dangling IDREFs (Dialog, AlertDialog, Progress)

sridhar-3009:
## Problem

`DialogContent` and `AlertDialogContent` unconditionally emit `aria-labelledby={titleId}` and `aria-describedby={descriptionId}` even when `DialogTitle` / `DialogDescription` are not rendered. If the referenced element doesn't exist in the DOM, the IDREF is dangling — invalid per the ARIA spec. Screen readers may announce empty text or skip the dialog's accessible name.

Closes #608.

## Solution

Presence-track `Title` and `Description` sub-components using `ShallowRef<boolean>` in the root context, set/unset via `onMounted`/`onUnmounted`. Content components then gate the ARIA attributes on whether the sub-component is actually present:

```ts
'aria-labelledby': context.titlePresent.value ? context.titleId : undefined,
'aria-describedby': context.descriptionPresent.value ? context.descriptionId : undefined,
```

Files changed: `DialogRoot.vue`, `DialogTitle.vue`, `DialogDescription.vue`, `DialogContent.vue`, `AlertDialogRoot.vue`, `AlertDialogTitle.vue`, `AlertDialogDescription.vue`, `AlertDialogContent.vue`.

### DIFF
--- a/.changeset/fix-dialog-dangling-aria-idrefs-608.md
+++ b/.changeset/fix-dialog-dangling-aria-idrefs-608.md
@@ -1,0 +1,7 @@
+---
+"@vuetify/v0": patch
+---
+
+fix(Dialog,AlertDialog): omit `aria-labelledby`/`aria-describedby` when title/description are absent (#631)
+
+`DialogContent` and `AlertDialogContent` unconditionally emitted `aria-labelledby`/`aria-describedby` pointing at ids even when no `Title`/`Description` sub-component was rendered, producing a dangling IDREF — invalid per the ARIA spec and liable to make screen readers announce empty text or skip the dialog's accessible name. Both attributes are now gated on the referenced sub-component actually being mounted.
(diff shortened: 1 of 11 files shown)


johnleider:
Rebased onto current `master` (was 154 commits behind / conflicting).

Conflict resolution was mechanical — intended a11y behavior unchanged:
- Kept master's AlertDialog `cancelEl` / `focusCancel` (open focuses Cancel)
- Applied presence-gated `aria-labelledby` / `aria-describedby` on top

Please verify:
- Title/description present → attributes point at the ids
- Title/description absent or dynamically removed → attributes omitted (no dangling IDREF)
- AlertDialog still focuses Cancel on open

PR #635, opened 2026-07-17
feat(Select): add label prop to SelectActivator

### LINKED ISSUE #613: [Bug] Provider Items ship a partial ARIA contract: aria-selected without a role, no keyboard path
From the #543 WS4 audit. The provider Item components emit state ARIA that is invalid or inoperable without consumer completion:

- SelectionItem.vue:104, SingleItem.vue:104, StepItem.vue:104 emit `aria-selected` with **no role** — the attr is only valid on option/tab/row/gridcell/treeitem, so it is ignored as emitted.
- All four Item families bind `onClick` but no `tabindex`/`onKeydown` (GroupItem.vue:118-126, SelectionItem.vue:103-108, SingleItem.vue:103-108, StepItem.vue:103-108) — keyboard users can't operate them unless the consumer adds the missing half.
- StepRoot exposes first/last/next/prev/step in slot props (StepRoot.vue:100-104) but nothing wires roving tabindex or arrow keys (useRovingFocus exists but isn't composed).

Design decision needed: either (a) complete the contract where the role is knowable (GroupItem is role=checkbox — tabindex+Space is a safe ~8 LOC addition), (b) document role+keyboard as a required consumer step for the role-agnostic families, or (c) accept a role prop that unlocks the full wiring. GroupRoot.vue:116 / SelectionRoot.vue:105 also emit aria-multiselectable on role-less containers (low).

sridhar-3009:
`SelectActivator` is a combobox trigger but had no way to carry an accessible label when the visible label is in a sibling element, and no typeahead for keyboard users. This closes both gaps.

## Changes

**`packages/0/src/components/Select/SelectActivator.vue`**
- New optional `label` prop → emitted as `aria-label` on the activator (passed through to `SelectActivatorProps` / `SelectActivatorSlotProps.attrs`)
- In `onKeydown`, when the listbox is open and a printable character key is pressed (single char, no Ctrl/Meta/Alt modifier), delegates to `context.virtualFocus.typeahead(e.key)` — jumps virtual focus to the first non-disabled item whose text content starts with that character

**`packages/0/src/composables/useVirtualFocus/index.ts`**
- New `typeahead(char: string) => void` method on `VirtualFocusReturn` (interface + implementation)
- Searches all non-disabled items starting from the position after the currently highlighted one (wrapping around), finds the first whose visible text starts with the typed char (case-insensitive), and calls `highlight()` on it

## WCAG coverage

- **2.1.1 Keyboard** — printable-char typeahead is required for listboxes per ARIA authoring practice (APG Listbox pattern §6.5)
- **4.1.2 Name, Role, Value** — `aria-label` on the combobox satisfies the accessible name requirement when the label element is not associated via `<label for>`

Refs #613 (already closed by #633; this PR completes the Select half of the same audit).


johnleider:
This PR has been split. The printable-char typeahead was removed in 9ab7835fe and now lives as a design item in #908: the `typeahead` name breaks `VirtualFocusReturn`'s imperative verb family, its placement is contested (`useVirtualFocus` vs a shared text-matching seam that `useRovingFocus` could also consume), and the implementation was partial APG — no timed multi-character buffer, and inert while the listbox is closed where native `<select>` responds. The PR now carries only the `SelectActivator` `label` prop, which resolves the accessible-name gap on its own and is mergeable standalone. The typeahead prototype remains fully recoverable at the pre-split head, bdb78da8e. Thanks for the contribution — the typeahead work isn't lost, it just needs the design pass before freezing onto the composable API.

PR #636, opened 2026-07-18
fix(Toggle): add Enter key support and role=button for non-native as elements

### LINKED ISSUE #616: [Bug] Non-native `as` elements get button ARIA but dead keyboard (Button, Toggle, PaginationItem) + Atom type=button default
From the #543 WS4 audit — a cross-family sweep item:

- ButtonRoot.vue:210-226: `as='div'` renders role=button + tabindex but no Enter/Space onKeydown.
- ToggleRoot.vue:149-154: handles Space only; Enter relies on native button semantics; non-native `as` also never gets role=button (:173).
- PaginationItem.vue:112-122: same native-button reliance.
- Atom.vue:101-108: `as="button"` doesn't default `type="button"` — a bare polymorphic button inside a form submits it. ~4 LOC but a behavior change; decide intentionality.

Fix pattern: when `as` is not a native button, add role=button (where missing), and an onKeydown handler mapping Enter/Space to the click path. Sibling components (CollapsibleActivator.vue:84-104, ExpansionPanelActivator.vue:89-110) already implement exactly this fallback — mirror them.

sridhar-3009:
## Problem

`ToggleRoot` has two related accessibility gaps when rendered as a non-native element (`as="div"`, `as="span"`):

1. **Missing Enter activation** — `onKeydown` only handled `Space`. Native `<button>` converts Enter to a click automatically; `<div role="button">` does not. Keyboard-only users on a non-native Toggle can focus it but cannot activate it with Enter.

2. **Missing `role="button"`** — the `aria-pressed` state attribute implies `role="button"` semantically, but many AT rely on an explicit role to announce the control correctly. Without it, some screen readers announce the element's implicit tag role (generic/group) instead of \"button\".

These were identified in the WS4 accessibility audit (closes #616, ToggleRoot items).

## Solution

**`onKeydown`**: extend the guard to accept both `Space` and `Enter`:
```ts
// before
if (e.key !== ' ') return
// after
if (e.key !== ' ' && e.key !== 'Enter') return
```

`e.preventDefault()` is called for both keys. For native `<button>` this prevents the browser from generating a redundant click event on Space/Enter; for non-native elements there is no native event to prevent, so it is a no-op that does no harm.

**`role`**: added `'role': as === 'button' ? undefined : 'button'` to `slotProps.attrs`. Native `<button>` already has its implicit button role; emitting `role="button"` on it would be redundant. The TypeScript interface is updated to include the new field.

## Tests

- Updated the existing `should toggle on Enter key via native button behavior` test to directly trigger `keydown Enter` (previously triggered `click` as a workaround since Enter wasn't wired).
- Added `should have role=button when rendered as non-native element`
- Added `should not have role attribute on native button element`
- Added `should toggle on Enter key when rendered as non-native element`
- Added `should toggle on Space key when rendered as non-native element`

All 6373 tests pass.

### DIFF
--- a/.changeset/fix-toggle-enter-role-616.md
+++ b/.changeset/fix-toggle-enter-role-616.md
@@ -1,0 +1,7 @@
+---
+"@vuetify/v0": patch
+---
+
+fix(Toggle): add Enter key support and `role="button"` for non-native `as` elements (#636)
+
+`ToggleRoot`'s `onKeydown` only handled Space, so a `Toggle` rendered as a non-native element (`as="div"`, `as="span"`) could be focused but not activated with Enter. It also never emitted `role="button"` for non-native elements, so some assistive technology announced the implicit tag role instead. Both are fixed: Enter now toggles alongside Space, and `role="button"` is set whenever `as !== 'button'`.
--- a/packages/0/src/components/Toggle/ToggleRoot.vue
+++ b/packages/0/src/components/Toggle/ToggleRoot.vue
@@ -53,6 +53,7 @@
     toggle: () => void
     /** Attributes to bind to the toggle element */
     attrs: {
+      'role': 'button' | undefined
       'type': 'button' | undefined
       'aria-pressed': boolean
       'aria-disabled': boolean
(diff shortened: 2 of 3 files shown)


sridhar-3009:

PR #638, opened 2026-07-18
fix(Progress): only emit aria-labelledby when Progress.Label is mounted

### LINKED ISSUE #608: [Bug] a11y: unconditional aria-labelledby/describedby emit dangling IDREFs (Dialog, AlertDialog, Progress)

sridhar-3009:
## Summary

- `ProgressRoot` unconditionally emitted `aria-labelledby` pointing to `${id}-label` even when no `Progress.Label` child was mounted, creating a dangling IDREF that screen readers announce incorrectly
- Adds `hasLabel: ShallowRef<boolean>` to `ProgressRootContext`; `ProgressLabel` sets it `true` on mount and `false` on unmount, so `aria-labelledby` is only emitted when a visible label is actually present
- Adds `ariaLabel` string prop to `ProgressRoot` for cases where no visible label exists but an accessible name is still needed

## What changed

- `ProgressRoot.vue`: adds `hasLabel` ref to context; `aria-labelledby` is conditional on `hasLabel.value`; new `ariaLabel` prop exposed as `aria-label` when no label is mounted
- `ProgressLabel.vue`: `onMounted` / `onBeforeUnmount` lifecycle hooks update `root.hasLabel`
- `index.test.ts`: updated one existing test that assumed unconditional `aria-labelledby`; added four new tests covering the no-label, with-label, `ariaLabel` prop, and precedence cases

## Test plan

- [ ] `aria-labelledby` is absent when `Progress.Label` is not mounted
- [ ] `aria-labelledby` is set when `Progress.Label` is mounted
- [ ] `aria-label` is set from `ariaLabel` prop when no `Progress.Label` is present
- [ ] `aria-labelledby` wins over `aria-label` when both are present
- [ ] All 178 test files pass (`pnpm test`)

Closes #608

### DIFF
--- a/.changeset/fix-progress-aria-labelledby-608.md
+++ b/.changeset/fix-progress-aria-labelledby-608.md
@@ -1,0 +1,7 @@
+---
+"@vuetify/v0": patch
+---
+
+fix(Progress): only emit `aria-labelledby` when `Progress.Label` is mounted (#638)
+
+`Progress.Root` unconditionally emitted `aria-labelledby` pointing to a label id even when no `Progress.Label` child was mounted, creating a dangling IDREF that assistive technology could misreport. `aria-labelledby` is now conditional on a label actually being present, and a new `ariaLabel` prop covers the case where no visible label exists but an accessible name is still needed.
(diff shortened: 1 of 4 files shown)


sridhar-3009:

johnleider (approved):

PR #639, opened 2026-07-18
fix(Carousel): add onKeydown to Next/Previous for non-native element keyboard activation

sridhar-3009:
## Summary

- `Carousel.Next` and `Carousel.Previous` only exposed `onClick` in their `attrs` slot props. When rendered with `as='div'` (or any non-button element), browsers do not synthesize `click` events from Enter or Space keypresses, silently breaking keyboard navigation for users who rely on keyboards or AT
- Adds `onKeydown` handlers to both components that call the existing `onClick` logic when Enter or Space is pressed — the same pattern already used by `CarouselIndicator`
- The `onKeydown` type is added to `CarouselNextSlotProps.attrs` and `CarouselPreviousSlotProps.attrs` so TypeScript consumers know it's available

## What changed

- `CarouselNext.vue`: new `onKeydown` function calling `onClick()` on Enter/Space; exposed in `slotProps.attrs`; type added to `CarouselNextSlotProps`
- `CarouselPrevious.vue`: same changes symmetrically
- `index.test.ts`: two new tests — one for each component — mounting with `as='div'` and asserting Enter/Space trigger slide navigation

## Test plan

- [ ] `Carousel.Next` rendered as `div`: Enter key advances to next slide
- [ ] `Carousel.Previous` rendered as `div`: Enter key moves to previous slide
- [ ] Space key also triggers navigation (and `defaultPrevented`)
- [ ] All 100 Carousel tests pass

### DIFF
--- a/.changeset/fix-carousel-next-prev-keydown.md
+++ b/.changeset/fix-carousel-next-prev-keydown.md
@@ -1,0 +1,7 @@
+---
+"@vuetify/v0": patch
+---
+
+fix(Carousel): add `onKeydown` to `Next`/`Previous` for non-native element keyboard activation (#639)
+
+`Carousel.Next` and `Carousel.Previous` only exposed `onClick` in their `attrs` slot props. When rendered with `as="div"` (or any non-button element), browsers do not synthesize click events from Enter or Space, silently breaking keyboard navigation. Both now wire an `onKeydown` handler that calls the existing click logic on Enter/Space, matching the pattern already used by `CarouselIndicator`.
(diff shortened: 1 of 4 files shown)


johnleider:
Superseded by the full host-polyfill sweep in the PR that lands role + tabindex + gated Enter/Space (not keydown-only). See `fix/button-host-polyfill`.

johnleider:
Closing as superseded by https://github.com/vuetifyjs/0/pull/751 (full non-button host polyfill: role + tabindex + gated Enter/Space across the default-button family, not keydown-only).

PR #640, opened 2026-07-18
fix(NumberField): add labelledBy prop for aria-labelledby on spinbutton

sridhar-3009:
## Summary

- `NumberField.Control` only supports accessible naming via the `label` string prop (rendered as `aria-label`). Consumers who render a visible `<label>` element outside the component cannot wire it to the spinbutton, and omitting `label` silently produces a spinbutton with no accessible name — a WCAG 4.1.2 failure
- Adds `labelledBy` string prop to `NumberField.Root` that flows through context to `NumberField.Control` and renders as `aria-labelledby` on the spinbutton
- When `labelledBy` is set, `aria-label` is suppressed to avoid redundant or conflicting accessible names

## What changed

- `NumberFieldRoot.vue`: `labelledBy?: string` added to `NumberFieldRootProps` and `NumberFieldRootContext`; destructured in `defineProps` and threaded through context
- `NumberFieldControl.vue`: `aria-label` is now conditional on `labelledBy` being absent; `aria-labelledby` set from `root.labelledBy` when present
- `index.test.ts`: three new tests — `aria-labelledby` set from prop, `aria-label` suppressed when `labelledBy` is present, `aria-label` absent when neither prop is provided

## Test plan

- [ ] `labelledBy="my-label"` → `aria-labelledby="my-label"` on the spinbutton
- [ ] `label="Quantity" labelledBy="my-label"` → `aria-labelledby` wins, no `aria-label`
- [ ] No props → no `aria-label` on spinbutton
- [ ] All 138 NumberField tests pass

Fixes a WCAG 4.1.2 gap where consumers with visible external labels had no way to wire them to the spinbutton.


sridhar-3009:

johnleider (approved):

PR #645, opened 2026-07-18
fix: add keyboard activation for non-native as elements (Button, Toggle, PaginationItem)

### LINKED ISSUE #616: [Bug] Non-native `as` elements get button ARIA but dead keyboard (Button, Toggle, PaginationItem) + Atom type=button default

sridhar-3009:
Closes #616.

## Problem

`ButtonRoot`, `ToggleRoot`, and `PaginationItem` all expose `role=button` and `tabindex` when rendered with a non-native `as` element (e.g. `as="div"`), but none had an `onKeydown` handler. Browsers only synthesize a `click` event from Enter/Space for native `<button>` elements — for all other elements, Enter and Space are dead keys.

## Fix

Mirror the pattern already used in `CollapsibleActivator.vue` and `ExpansionPanelActivator.vue`:

- **ButtonRoot**: add `onKeydown` (Enter/Space → `onClick`); only exposed when `as !== 'button'`
- **ToggleRoot**: extend the existing Space-only `onKeydown` to also handle Enter; add `role=button` when `as !== 'button'` (it was missing entirely — only `type` was gated)
- **PaginationItem**: add `onKeydown` (Enter/Space → `select`); add `role=button` when `as !== 'button'`

## Tests

Added 3 tests per component (9 total) covering:
- Enter key activates when `as='div'`
- Space key activates when `as='div'`
- `onKeydown` / `role` not exposed when rendered as a native `<button>`

### DIFF
--- a/.changeset/fix-non-native-keyboard-activation-616.md
+++ b/.changeset/fix-non-native-keyboard-activation-616.md
@@ -1,0 +1,7 @@
+---
+"@vuetify/v0": patch
+---
+
+fix(Button,Toggle,Pagination): activate non-native `as` elements on Enter/Space (#645)
+
+`ButtonRoot`, `ToggleRoot`, and `PaginationItem` expose `role="button"` and `tabindex` when rendered with a non-native `as` element (e.g. `as="div"`), but browsers only synthesize a click from Enter/Space for native `<button>` elements — for everything else those were dead keys. All three now wire an `onKeydown` handler (only when `as !== 'button'`) so keyboard users can activate them.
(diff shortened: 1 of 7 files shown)


sridhar-3009:

PR #646, opened 2026-07-19
fix: add role prop to SnackbarRoot for alert vs status (#615)

### LINKED ISSUE #615: [Bug] Snackbar: no alert-role path and unreliable first announcement
From the #543 WS4 audit. Two related live-region gaps:

1. SnackbarRoot.vue:81 hardcodes `role=status` while the header comment (:8) promises "alert for urgent, status for informational" — no prop exists to choose. (~5 LOC prop.)
2. The live-region node mounts with its text already inside, instead of the empty-then-fill pattern; many SRs miss the first announcement. PaginationStatus.vue:74-106 implements the correct deferred pattern (100ms empty-then-fill) to mirror.

Focus is correctly never stolen (non-modal portal, scrim:false SnackbarPortal.vue:67) and the close button is labeled (SnackbarClose.vue:61-62).

sridhar-3009:
## Summary

Closes #615.

`SnackbarRoot` hardcoded `role="status"` on every snackbar. This makes it impossible for callers to announce urgent messages correctly — screen readers treat `role="status"` as a polite live region, but destructive or time-sensitive notifications should use `role="alert"` (assertive).

### Changes

- **`SnackbarRoot.vue`** — adds a `role?: 'alert' | 'status'` prop (defaults to `'status'` to preserve existing behaviour). The prop is forwarded through both the element's DOM attribute and `slotProps.attrs.role` so renderless consumers can spread it onto their own element.
- **`index.browser.test.ts`** — two new tests: default is `'status'`; setting `role="alert"` propagates to the DOM and slot props.

### Before / After

```vue
<!-- before: only status, no override possible -->
<Snackbar.Root>...</Snackbar.Root>

<!-- after: opt-in to alert for urgent messages -->
<Snackbar.Root role="alert">...</Snackbar.Root>
```

### Test plan

- [x] `should default to role="status"` — no prop → DOM has `role="status"`
- [x] `should set role="alert" when role prop is alert` — explicit prop propagates to DOM and `slotProps.attrs.role`
- [x] Pre-existing `should pass role attribute through to element` test still passes (unaffected)
- [x] Full test suite passes (`pnpm test:run --project '!v0:browser'`)

### DIFF
--- a/packages/0/src/components/Snackbar/SnackbarRoot.vue
+++ b/packages/0/src/components/Snackbar/SnackbarRoot.vue
@@ -40,12 +40,14 @@
     namespace?: string
     /** Unique identifier. Auto-generated if not provided. */
     id?: ID
+    /** ARIA live-region role. Use 'alert' for urgent messages, 'status' for informational. @default 'status' */
+    role?: 'alert' | 'status'
   }
 
   export interface SnackbarRootSlotProps {
     id: ID
     attrs: {
-      role: string
+      role: 'alert' | 'status'
     }
   }
 </script>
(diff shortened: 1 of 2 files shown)


johnleider:
Closing as superseded. The team went with the `urgent` boolean approach in #624 (now merged) rather than exposing a raw `role` string prop — a boolean keeps the ARIA detail internal to the headless component. Thanks for the contribution here; the underlying issue (#615) is fixed for the alert-role facet by #624.

PR #681, opened 2026-07-22
docs: dot grid skew (spherical curvature) setting

johnleider:
## Summary

Adds a **Skew** control to Settings → Background Effects that curves the docs dot grid onto a sphere — dots spread apart at the pole and compress toward the rim, like a flat sheet wrapping over a ball. The slider runs `-100`–`100` (0 = flat; sign flips convex/concave).

### Approach

Earlier iterations tried a uniform axis-stretch and a CSS perspective tilt — both are planar transforms and read as flat/tilted, not spherical. A real sphere needs a *position-dependent* warp, so this renders on a canvas:

- **`AppDotSphere.vue`** (new, docs-only) — a canvas that projects each flat grid node through a spherical mapping (distance from centre → arc angle, so equal steps compress toward the rim), connects neighbours into a warped lat/long mesh, and shrinks/dims dots by surface depth. Sizes via `useResizeObserver`; redraws on prop/theme change. SSR-safe (draws on mount only).
- The flat CSS grid stays the default. The canvas mounts **only when skewed**, so the 99% who never touch the setting pay nothing (no canvas, no redraw).
- No `@paper/genesis` change — the genesis `GnDotGrid` primitive is untouched, so there's no DS version bump. The feature is entirely in `apps/docs`.

Wiring:

- `useSettings` — persisted `dotGridSkew` (default `0`), in `hasChanges`/`reset`.
- `App.vue` — mounts `AppDotSphere` inside the shell (above the translucent shell background, below content) when `showDotGrid && dotGridSkew !== 0`; keeps the CSS `.dot-grid` for the flat case.
- `AppSettingsTheme` — `AppSettingsSlider` (`-100`..`100`) with a live sphere preview swatch.

## Verification

- `pnpm typecheck` (packages) — green; genesis reverted to master (net-zero).
- Live in browser: canvas paints the spherical dot field (confirmed via pixel sampling + preview swatch); flat default (skew 0) uses the CSS grid with no canvas mounted; toggling to a nonzero value swaps in the sphere.

Defaults to `0`, so the out-of-the-box appearance is unchanged.

### DIFF
--- a/.changeset/gn-dot-grid-skew.md
+++ b/.changeset/gn-dot-grid-skew.md
@@ -2,4 +2,4 @@
 "@paper/genesis": minor
 ---
 
-feat(GnDotGrid): add `skew` prop for directional grid stretch — a `-100`–`100` value that stretches the dot grid along one axis while pinching the other by an equal amount, so the pattern reads as pulled in a direction. Defaults to `0` (square cells), so existing usage is unchanged.
+feat(GnDotGrid): add `skew` prop for perspective grid tilt — a `-100`–`100` value that bends the flat dot grid into a receding surface so the dots bunch toward the far edge, with the sign choosing which edge tips away. Defaults to `0` (flat), so existing usage is unchanged.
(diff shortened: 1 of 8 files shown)


johnleider:
Closing — abandoning this approach per author; not merging.

PR #699, opened 2026-07-22
chore: self-host the coverage badge, dropping the Codecov dependency

johnleider:
## Problem

The README coverage shield renders **"unknown"** — and it's **not** our CI. The upload step succeeds every run (`Upload queued for processing complete`, token present), but Codecov has produced **no report for any master commit today**: the last 30 commits are all `state/totals/report: null` via Codecov's API. The shields badge reads the HEAD commit's report → none → "unknown". (The ~99% still visible via the API is a stale cached total.) Codecov-side processing, unfixable from the repo — so self-host the badge.

## How

- `scripts/lib/coverage-badge.js` — computes overall statement coverage for the published package (`packages/0/src`, minus `.test.`/`.bench.`) from `coverage-final.json`; writes a shields.io **endpoint** badge JSON (Codecov-matched color thresholds). Writes nothing when there's no source coverage (stale badge beats "0%").
- `scripts/generate-metrics.js` — emits `apps/docs/public/coverage-badge.json` after the coverage pass it already runs (no extra test run).
- `.github/workflows/metrics-regen.yml` — commits the badge via the existing auto-PR.
- `README` — coverage shield now uses `img.shields.io/endpoint?url=…/master/apps/docs/public/coverage-badge.json`.
- Seeded at **98%** so it renders on merge; pipeline refreshes each regen.

## Notes

- Statement coverage (Codecov reported line, ~99% — same ballpark). Refreshes on regen cadence, not per-commit.
- Codecov uploads in `pr-checks.yml` left in place (harmless).
- Doesn't touch `metrics.json`/benchmark data.

## Test plan
- [x] `eslint` + `node --check` clean on both scripts
- [x] Seed produced valid shields-endpoint JSON (98%, brightgreen)
- [ ] After merge: shield renders from the raw endpoint


johnleider:
Superseding this workaround. Root cause found: PR #663 bumped `@vitest/coverage-v8` 4.1.5→4.1.10, whose `coverage-final.json` Codecov 5 can't process (confirmed via Codecov API — the parent of #663 reports complete/99.17%, #663 and everything after report null). Fixing at the source instead by uploading lcov, which Codecov parses reliably.

PR #734, opened 2026-07-28
chore(bench): record why BASELINE_ANCHOR_HZ must not be captured on GHA

johnleider:
## What

Comment-and-warning-text only. `BASELINE_ANCHOR_HZ` stays `null`, `computeScale` still returns 1, no artifact changes.

## Why

The doc comment on `BASELINE_ANCHOR_HZ` currently reads:

> To capture: run metrics-regen, read `apparatus.anchors` out of the produced benchmarks.json, paste it here…

Followed literally, that arms v0's benchmark unit off a GitHub Actions runner — which `.claude/specs/2026-07-24-dedicated-bench-host.md` §6 had already ruled out three days ago ("the in-flight GHA median-of-three capture was **halted rather than committed**"). The instruction and the decision disagree, and the instruction is the one a maintainer reads at the moment of temptation. `metrics:check`'s warning had the same gap — it named the constant but not the host.

Evidence the spec establishes, now cited at the constant:

- `runs-on: ubuntu-24.04` pins the image, not the CPU; the pool spans ~1.9x single-thread throughput.
- At least two vendors, and they do not differ by a scalar: 0.55–0.86 per-anchor Xeon 6973P-C → EPYC 7763. One scale factor cannot describe that.
- A median of three GHA runs does not help — with two vendors in three draws, the median selects the majority vendor, not a typical host.

Measured here on the committed artifacts, GHA also fails the spec's own acceptance bar (±2% aggregate, ±5% worst anchor) against *itself* — same job, same host, same runner image:

| comparison | trimmed gmean | per-anchor spread | worst anchor |
|---|---:|---:|---:|
| `1.0.0.json` (runs:1) vs `benchmarks.json` (runs:3) | 0.9909 | 0.916–1.015 | a06 map build 0.916 |
| `1.0.1.json` (runs:1) vs `benchmarks.json` (runs:3) | 0.9151 | 0.776–1.016 | a09 string build 0.776 |
| `1.0.1` vs `1.0.0` (both runs:1) | 0.9274 | 0.805–1.006 | a12 grouped aggregate 0.805 |

Part of that is 1-run vs 3-run median noise rather than host disagreement — but `metrics:history` snapshots *are* `runs:1`, so it is the noise the baseline would actually be asked to correct.

## Notes

- No changeset: `scripts/` only, and `changeset-reminder.js` gates on `packages/*/src/`. Ships no package version → `master` train, `chore` type.
- Verified: `node scripts/check-benchmark-artifacts.ts` exits 0 and emits the amended warning; eslint clean on both files.

### DIFF
--- a/scripts/check-benchmark-artifacts.ts
+++ b/scripts/check-benchmark-artifacts.ts
@@ -108,7 +108,8 @@ function main (): void {
       console.warn(
         '[check-benchmark-artifacts] warning: no calibration baseline stored yet — numbers are raw. '
         + 'Capture `apparatus.anchors` into BASELINE_ANCHOR_HZ (scripts/lib/calibration.ts) to enable '
-        + 'host normalization.',
+        + 'host normalization — on the dedicated reference host, never from a GHA run '
+        + '(see .claude/specs/2026-07-24-dedicated-bench-host.md §6).',
       )
     }
   }
--- a/scripts/lib/calibration.ts
+++ b/scripts/lib/calibration.ts
@@ -52,22 +52,8 @@ export const ANCHOR_COUNT = 13
  * records `scale: 1, baseline: null`, so the pipeline behaves exactly as it
  * does today and nothing is silently rescaled.
  *
- * Do NOT capture this from a GHA run, however tempting a green metrics-regen
- * artifact looks. `runs-on: ubuntu-24.04` pins the image, not the CPU: the pool
- * spans ~1.9x in single-thread throughput across at least two vendors whose
- * anchors do not differ by a scalar (0.55–0.86 per-anchor, Xeon 6973P-C to EPYC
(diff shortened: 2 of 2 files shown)


johnleider:
Obsolete. This documents why `BASELINE_ANCHOR_HZ` must not be captured on a GHA runner; #749 deleted `BASELINE_ANCHOR_HZ`, `computeScale`, and `scripts/lib/calibration.ts` entirely, because measurement showed the normalization made results worse rather than better (per-bench disagreement on identical code: 1.81% raw vs 3.53% normalized).

The concern behind this PR was right and is now enforced structurally instead of by comment: benchmarks are measured only on the fixed reference host, CI fails if it touches `benchmarks.json`, and the reasoning is recorded under "Why there is no host calibration" in `.claude/rules/benchmarks.md`.

PR #749, opened 2026-07-29
chore(bench): delete host calibration; fixed host, raw numbers, verifiable

johnleider:
Benchmarks were unreliable because GHA rotates CPUs. This fixes that by fixing the host — and **deletes** the calibration apparatus that was built to compensate for it, because measurement shows it made things worse.

## The headline

The apparatus doubled the error it existed to remove. Four full-suite runs of identical code on the fixed host, 433 real benches, 6 run-pairs (2,598 comparisons):

| method | per-bench median | p95 | >5% | per-feature median | per-feature worst |
|---|---|---|---|---|---|
| **raw ops/s** | **1.81%** | **8.51%** | **366** | **2.16%** | **4.96%** |
| anchor-normalized *(was shipping)* | 3.53% | 10.46% | 853 | 6.23% | 9.50% |

Bootstrap 95% CI on the regression: **+1.72pp [1.52, 1.88]** — significant, not noise. Tier flips on identical code: 7 raw vs 10 normalized.

**Why it could never have worked.** Decomposing pairwise variance, the run-wide common shift accounts for **0.1–13%**; 87–99.9% is per-bench noise no global scalar can touch. There is ~1% of host drift available to correct against a ~4% noise floor. On one of four runs the anchors read −4.6% median while the real suite ran **+1.44% faster** — wrong sign — and dividing by 0.9586 injected +4.32% into every published number. Trimming can't rescue it (TRIM 0/1/2/3 → 0.914 / 0.939 / 0.959 / 0.961). The least reproducible file in the whole suite was `calibration.bench.ts` itself, 9.42% against a worst real feature of 4.96%.

**Two facts made deletion safe and obvious.** Every committed artifact carried `scale: 1` — the apparatus was inert its entire life, so no published number changes and there is no migration. And of ~35 comparable projects surveyed (V8, JetStream, Speedometer, Node core, TypeScript `ts-perf`, rustc-perf, LNT, SPEC, JMH, Go `benchstat`, criterion.rs, Vue core, React, Deno, Bun, esbuild, Biome, tinybench, mitata, CodSpeed, BenchmarkDotNet, MongoDB, Mozilla), **zero** derive a host scale from a synthetic probe and divide by it. Benchmark.js shipped exactly this and removed it for exactly this reason. SPEC's reference machine is not precedent — it normalizes each benchmark against *its own* reference time, which provably cancels.

## The process now

1. **One fixed workstation.** Removes rotation rather than correcting for it. This is the whole mechanism, and the part with real precedent (rustc-perf, `ts-perf`, Chrome's device lab).
2. **`--runs 3`, median-merged.** Repeats are the only lever that measurably works: 1 run/side → 1.95% median error, 2 runs/side → **1.20%**. Larger than every normalizer combined, and already in place.
3. **Record, don't correct.** `env` fingerprint (cpu/node/busy/peak/governor) travels with every artifact. Mismatched `cpu`/`node` means not comparable → re-measure, never rescale.
4. **Make host movement visible.** `suiteShiftPct` reports whole-suite median movement as one line, so common-mode drift is something a human reads rather than something silently folded into every figure.
5. **Refuse to measure a busy machine.** The contention guard stays (5% aggregate / 35% per-core, `iowait` counted as busy). Idle floor is 0.03–0.22% against a 5% limit, so false refusals are ~impossible and the cost of one is a re-run.
6. **`pnpm metrics:verify`.** The claim "these numbers are trustworthy" is now checkable rather than asserted — it measures identical code repeatedly and fails if results disagree beyond measured thresholds. `--runs N` to measure fresh, `--from a.json b.json` to re-check existing artifacts.

## Rejected, with reasons

- **Self-normalization** (divide by the run's own geomean). Marginally better on paper, catastrophic in practice: a −15% registry regression touching 306 of 433 benches reports as −4.7% *plus a phantom +12.2% speedup* on the untouched ones. Per-file geomean reports **0.0%** and never fires — it divides a composable's benches by that composable's own mean, so it structurally cannot detect a regression in the thing it gates. Fails silently toward "everything is fine".
- **Governor toggle.** Deleted. Its only evidence was CV 1.79% vs 1.98% — measured *on the anchors*, the component since proven not to track the suite. A root sudoers grant and signal-restore plumbing for a +0.06% effect measured with a discredited instrument. `readGovernor` stays to record and warn.
- **rme-aware gating.** `log(rme)` vs `log(cross-run spread)` correlates 0.413 — explains 17%. rme is within-run dispersion and does not predict between-run dispersion.
- **Core pinning.** Measures worse; `taskset` is not part of the apparatus.

## Honest limitations

- **A uniform regression is only visible to a human.** A Node bump or a dropped `/*#__PURE__*/` that slows everything equally shows as one `suiteShiftPct` line, not a gate trip. Actionable floor ~3%. No scheme tested closes this — self-normalizers make it structurally invisible, and the anchors mis-estimated it by 4.3% in the one run where it mattered.
- **Nine benches are unreliable regardless** (spread >20% on identical code, worst 46.5%), clustered in `createNested` and `useDate`. They are the entire >20% tail. Named in the rules doc; they look like bench bugs, not host noise.
- **Slow drift on the same physical machine is undetectable** — thermal degradation or a silent microcode update leaves the fingerprint identical. `metrics:verify` is the periodic check against this.

## Reference-host state

The `v0-governor` sudoers grant installed for the deleted toggle is now unused. It stays: narrow (two literal arguments, deliberately un-escalatable), audited, and useful for manual governor changes during operator work on the box. Nothing calls it, and nothing needs to.

## Verification

- `pnpm typecheck` green; `pnpm lint` clean; **139 test files / 4657 tests pass**.
- `pnpm metrics:verify` run against four real artifacts and against three fresh full-suite runs on the reference host.
- Net ~860 lines removed, ~460 added — the additions being the fingerprint and the verifier.


johnleider:
## Second-model review round (Codex GPT-5.5 + Grok 4.5)

Six findings, all reproduced against source or on the reference host before being accepted. Fixed in `e5a80e1`.

1. **Guard was blind to single-core contention.** Aggregate `/proc/stat` only — one saturated core on a 36-thread host is 2.8% of total CPU, under any whole-machine threshold, while being the likeliest real contention there is. Reproduced: one busy thread read 2.8% and passed. Now samples per-CPU and blocks above 35%, a limit set from the measured gap between idle (0–1% peak core) and a real workload (100%).
2. **Governor restore was unsound for any original but `powersave`.** `readGovernor()` can return `schedutil` / `ondemand` / `mixed(...)`, which the helper rejects, so restore failed and left the desktop pinned to `performance`. The hold now declines to touch an unrestorable governor.
3. **`mkdtempSync` sat between the governor change and the `try`** — a throw leaked `performance`.
4. **`finally` does not run on Ctrl-C.** Binding SIGINT/SIGTERM alone did nothing, because handlers only run when the event loop turns and `execFileSync` blocked it for the whole suite. Verified: SIGINT left vitest benching on and the box pinned, the interrupt queued behind the run it was meant to abort. `runOnce` now spawns async and the handler kills the child; re-verified the box returns to `powersave`.
5. **`CI=true` skip disabled the guard on the machine it was written for**, since the reference host under automation also sets `CI`. Ephemeral is now `CI` set **and** no governor helper. The host is always inspected regardless, so `apparatus.env` records conditions behind every artifact.
6. **`iowait` counted as idle.** Verified: a `dd` loop read 2.9% aggregate — invisible — and 38.4% on one core.

## Re-baseline, and a finding that changes how to read this PR

`BASELINE_ANCHOR_HZ` is now the per-anchor median of **four** full-suite runs, not one capture.

More important than the new numbers: across those four runs on the same idle host, three produced a scale within 0.4% of 1 while the fourth produced **0.963** — its allocation-heavy anchors collapsed (`object churn` −38%, `grouped aggregate` −27%) while the 446 real benches moved only **+1.4%**. The median anchor moved −4.6%, so no trim setting rescues it. Normalizing by that scale would have **injected ~4% of error** into every bench rather than removing any.

The async-spawn change landed between runs and was ruled out as the cause: a fourth run on the new code behaved normally (scale 0.9957).

**Honest reading:** on an unchanged host this apparatus is a no-op two runs in three and a ~4% distortion in the third. Its value is cross-host — insurance against a future CPU/kernel/Node change — and because every artifact stores raw anchors, that correction stays derivable retroactively. A sub-5% normalized delta should be read as noise. Both the code and the rules doc now say this.

This also retires the earlier "scale reproducible to 0.012%" claim from the PR body — that run was luckier than it was representative.

johnleider:
## Verified on the reference host

`pnpm metrics:verify` — two fresh median-of-3 samples of identical code, on this branch with calibration deleted:

| metric | measured | limit |
|---|---|---|
| feature median | **0.79%** | 5% |
| feature max | **2.42%** | 10% |
| bench median | **2.14%** | 6% |
| bench p95 | **11.55%** | 20% |

433 benches across 16 features. `PASS`.

For contrast, the same check against the four historical artifacts (which still contain the anchors): feature median 2.16%, bench median 3.93% — and the least reproducible file in that set was `calibration.bench.ts` itself at 9.42%, against a worst real feature of 4.96%.

### The verifier caught its own bug first

Its initial run **failed**: per-bench median 6.00% against a 4% band. That was correct behaviour and my error — it measured each sample with `--runs 1` while the bands were calibrated against the median-of-3 artifacts the pipeline actually publishes. Fixed in `c06f23e`: `--merge` defaults to 3 and warns when set otherwise.

That failure is the point of the tool. "These numbers are trustworthy" is now a claim anyone can check in one command, rather than something asserted in a PR description.

### Full verification

- `pnpm typecheck` green, `pnpm lint` clean, **139 test files / 4657 tests pass**
- `pnpm metrics:verify` passes against both fresh runs and historical artifacts
- Reference host left idle on `powersave`, no stray processes

johnleider:
## Gap found and fixed: CI was still measuring benchmarks

`metrics-regen.yml` ran `pnpm metrics` and `pnpm metrics:history --force` on `runs-on: ubuntu-24.04`. With calibration deleted, merging that combination would have been **worse than what preceded it** — a shared runner does not pin the CPU, GHA rotates hosts across generations differing ~1.5x, and nothing would be compensating. The next release would have replaced reference-host numbers with whichever host the job drew.

Split by whether a number cares which CPU produced it:

| | where | why |
|---|---|---|
| **Coverage** | CI, automatic | Deterministic. A shared runner is fine. |
| **Benchmarks + history** | Reference host, manual | The machine *is* the apparatus. No GitHub runner is that machine. |

The workflow now runs `pnpm metrics:coverage`, commits only `metrics.json`, and **fails loudly if `benchmarks.json` changes** — `generate-metrics.js` reads that file from the working tree, so the job re-emits the committed reference-host numbers unchanged rather than measuring new ones.

Verified on the reference host: after `pnpm metrics:coverage`, `benchmarks.json` md5 is byte-identical and only `metrics.json` is modified.

No automation for the benchmark half is deliberate, and matches the prior art — rustc-perf and TypeScript's `ts-perf` both use dedicated hardware rather than general CI.

## The regeneration runbook

```bash
# On the reference workstation, from a clean checkout of master
git pull && pnpm install --frozen-lockfile

pnpm metrics                 # coverage + build + bench (median of 3) + metrics.json
pnpm metrics:history         # only when a published version has no snapshot yet

pnpm metrics:delta --prev <previous benchmarks.json> --next apps/docs/public/benchmarks.json
pnpm metrics:check
```

Commit as `chore(bench): regenerate metrics` — that prefix is what `metrics:check` accepts for a human-authored artifact change, so feature branches still cannot quietly ship laptop numbers.

Read the delta at feature level (identical code reproduces to ~2%); a single bench moving <10% is usually noise. If `suiteShiftPct` shows the whole suite moving >5% one way, suspect the machine and check `env` against the previous artifact.

**When the machine changes** — new CPU, kernel, Node major, or the box is replaced — run `pnpm metrics:verify --runs 2` to confirm the new setup reproduces itself, then `pnpm metrics:history --force` so every point shares one machine, and say so in the PR. No factor bridges that seam.

Full docs: "Regenerating metrics" in `.claude/rules/benchmarks.md`.

Issue #157, opened 2026-03-12
[Feature Request] breakpoints need precise width to avoid misalignment with CSS on zoomed pages


J-Sek:
Subject: a user who has large monitor and scales pages ~150% on a regular basis

On a certain width the demo in [vuetify#22710](https://github.com/vuetifyjs/vuetify/pull/22710) shows mismatch
- JS reports `lg`, but the grid and `v-container` dropped to `md`
  - (I have set 150% page zoom and resized the page slowly)
  - (not sure if it matters, but the 4K monitor already has OS-level 150% scaling enabled)

<img width="1568" height="1044" alt="Image" src="https://github.com/user-attachments/assets/76d469a6-a37c-47d8-8baf-c83b7233464f" />

Suggestion:
- use `document.body.getBoundingClientRect().width` instead of `window.innerWidth` (the latter gives rounded value)

Note:
- the problem is already present in Vuetify. v5 makes a room to introduce an improvement

PR #761, opened 2026-08-01
fix(useBreakpoints): resolve the initial breakpoint via matchMedia, not innerWidth

### LINKED ISSUE #730: [Bug] createBreakpoints: initial state uses innerWidth while update() uses matchMedia
### Problem

`createBreakpoints` resolves its **initial** state by comparing `window.innerWidth` against the threshold numbers, but `update()` resolves the **same** state through `window.matchMedia`. The two mechanisms do not agree at every viewport, so a consumer of the bare factory renders first paint under one mechanism and every subsequent frame under the other.

The plugin masks this by calling `update()` at setup. The bare `createBreakpoints()` factory does not — and that is the path Vuetify's `display` composable takes.

### Evidence

`packages/0/src/composables/useBreakpoints/index.ts` at `adaace94d`.

**Initial state — width comparison** (`:138-171`):

```ts
const initialWidth = ssr
  ? ssr.clientWidth
  : (IN_BROWSER ? window.innerWidth : 0)          // :138-140

let initialName: BreakpointName = 'xs'
for (let i = sorted.length - 1; i >= 0; i--) {
  if (initialWidth >= sorted[i]![1]) {            // :147
    initialName = sorted[i]![0]
    break
  }
}
const initialIndex = names.indexOf(initialName)
...
const isMobile = shallowRef(initialWidth < mb!)   // :157
```

**Updates — matchMedia** (`:173-197`):

```ts
// Use matchMedia for breakpoint detection to guarantee
// alignment with CSS media queries at any zoom level.
// Falls back to innerWidth comparison when matchMedia is unavailable.
let current: BreakpointName = 'xs'
for (let i = sorted.length - 1; i >= 0; i--) {
  const px = sorted[i]![1]
  if (SUPPORTS_MATCH_MEDIA ? window.matchMedia(`(min-width: ${px}px)`).matches : width.value >= px) {  // :185
    current = sorted[i]![0]
    break
  }
}
...
isMobile.value = isNumber(mb)
  ? (SUPPORTS_MATCH_MEDIA ? !window.matchMedia(`(min-width: ${mb}px)`).matches : width.value < mb)     // :196
  : index < names.indexOf(mobileBreakpoint as BreakpointName)
```

The initial block is the **pre-fix mechanism**. matchMedia was adopted precisely because width comparison misaligns with CSS — see #157 (JS reported `lg` while the grid rendered `md`, at 150% page zoom). That fix landed in `update()` only; the initialization path kept the old behavior.

**The plugin hides it** (`:277-280`):

```ts
setup: (context, app, _options) => {
  // In SSR mode, skip the synchronous update to avoid hydration mismatch.
  // The hydration watcher below will call update() after hydration completes.
  if (IN_BROWSER && !_options?.ssr) context.update()
```

Under SSR the hydration watcher (`:292-294`) corrects it after hydration. Either way, plugin consumers get a matchMedia-derived value before anything renders. Bare-factory consumers do not — nothing calls `update()` until the first `resize` event.

### Conditions to reproduce

A viewport whose CSS media-query width and `window.innerWidth` straddle a threshold. Two sources:

1. **Fractional zoom** — the #157 case, and the reason the matchMedia path exists. At non-100% zoom the rounded `innerWidth` and the media-query evaluation can land on opposite sides of a threshold.
2. **Scrollbar width** — `window.innerWidth` includes a classic (non-overlay) scrollbar; in Chrome and Firefox the CSS `width` media feature evaluates without it, a ~15px gap. *(Moderate confidence: browser-dependent, and a non-issue with macOS overlay scrollbars. The zoom case above is the verified one.)*

Under either, a direct `createBreakpoints()` consumer renders one breakpoint band on first paint, then silently flips to another on the first resize event — with no real change in band.

### Motivating consumer

`vuetifyjs/vuetify`'s `display` composable (branch `next`, `packages/vuetify/src/composables/display.ts:123-127`) calls the bare factory:

```ts
const breakpoint = createBreakpoints({
  mobileBreakpoint: options?.mobileBreakpoint,
  breakpoints: options?.thresholds,
  ssr: isObject(ssr) ? ssr : ssr === true ? { clientWidth: 0 } : undefined,
})
```

and `packages/vuetify/src/framework.ts:97` gates the corrective call on SSR:

```ts
if (IN_BROWSER && options.ssr) {
  // …app.$nuxt hook / app.mount wrapper → display.update()
}
```

So a non-SSR Vuetify app never corrects the initial values. `display.width` / `display.mobile` / `display.lgAndUp` stay width-comparison-derived until the user resizes the window — which is exactly the mode #157 was filed against.

### Fix sketch

Resolve the initial band through the same matchMedia path when `IN_BROWSER && !ssr` — extract the `:182-197` resolution into a shared helper and seed from it, or simply run `update()` at the end of `createBreakpoints()` under that condition. The SSR branch must keep seeding from `ssr.clientWidth` so hydration markup still matches.

### This is a behavior change

Consumers of the bare factory can see a different band at first paint after the fix. That is the correction (it will now match the band they get one resize later, and match CSS), but it can shift what renders on mount, and the release note should say so.

sridhar-3009:
Closes #730.

`createBreakpoints`'s initial state was resolved by comparing `window.innerWidth` against the breakpoint thresholds, while `update()` resolved the same state through `matchMedia` (adopted in the #157 fix). At non-100% zoom, or with a non-overlay scrollbar, the two can disagree on which side of a threshold the viewport falls — so a direct `createBreakpoints()` consumer (e.g. Vuetify's `display` composable outside SSR mode) could render one breakpoint band on first paint and silently flip to another on the first resize event, with no real viewport change in between.

The plugin's `setup` hook already called `context.update()` once at mount to paper over this, which is why the bug was invisible to plugin consumers — it only shows up for the bare factory.

## Fix

`createBreakpoints()` now runs `update()` once synchronously when `IN_BROWSER && !ssr`, so the initial band is resolved the same way as every subsequent one. The SSR branch is untouched and still seeds from `ssr.clientWidth` so hydration markup keeps matching.

## Tests

Added two cases to `index.test.ts`:
- initial breakpoint resolves via `matchMedia` when it disagrees with `innerWidth`
- SSR-seeded initial state is unaffected (still resolves from `ssr.clientWidth`, not `matchMedia`)

All 66 existing + new tests pass.

## Note

Per the issue, this is a behavior change for bare-factory, non-SSR consumers: the initial band can differ from before, though only in cases where it previously disagreed with the very next resize/update — i.e. it's a correction, not a regression.

### DIFF
--- a/.changeset/fix-breakpoints-initial-matchmedia-730.md
+++ b/.changeset/fix-breakpoints-initial-matchmedia-730.md
@@ -1,0 +1,16 @@
+---
+"@vuetify/v0": patch
+---
+
+fix(useBreakpoints): resolve the initial breakpoint via matchMedia, not innerWidth (#730)
+
+`createBreakpoints`'s initial state was resolved by comparing `window.innerWidth` against
+the breakpoint thresholds, while `update()` resolved the same state through `matchMedia`.
+At non-100% zoom or with a non-overlay scrollbar the two can disagree, so a direct
+`createBreakpoints()` consumer (e.g. Vuetify's `display` composable outside SSR mode)
+could render one breakpoint band on first paint and silently flip to another on the
+first resize event, with no real viewport change in between.
+
+`createBreakpoints()` now runs `update()` once synchronously when `IN_BROWSER && !ssr`,
+so the initial band is resolved the same way as every subsequent one. The SSR path is
+unchanged and still seeds from `ssr.clientWidth` to keep hydration markup matching.
--- a/packages/0/src/composables/useBreakpoints/index.test.ts
+++ b/packages/0/src/composables/useBreakpoints/index.test.ts
@@ -121,6 +121,35 @@ describe('createBreakpoints', () => {
         expect(context.xxl.value).toBe(false)
       })
 
+      it('should resolve the initial breakpoint via matchMedia, not innerWidth, when they disagree', () => {
+        // innerWidth says 'lg' (>= 1280), matchMedia says 'md' — simulates
+        // the fractional-zoom / scrollbar-width straddle described in #730.
+        mockWindow.innerWidth = 1280
+        mockWindow.matchMedia = (query: string) => {
+          const match = query.match(/\(min-width:\s*(\d+)px\)/)
+          const threshold = match ? Number(match[1]) : 0
+          return { matches: threshold <= 1024 }
+        }
+
+        const context = createBreakpoints()
+
+        expect(context.name.value).toBe('md')
+        expect(context.lg.value).toBe(false)
+        expect(context.md.value).toBe(true)
+      })
+
+      it('should not resolve the initial breakpoint via matchMedia when ssr dimensions are provided', () => {
+        mockWindow.matchMedia = (query: string) => {
+          const match = query.match(/\(min-width:\s*(\d+)px\)/)
+          const threshold = match ? Number(match[1]) : 0
+          return { matches: threshold <= 1024 }
+        }
+
+        const context = createBreakpoints({ ssr: { clientWidth: 1280 } })
+
+        expect(context.name.value).toBe('lg')
+      })
+
       it('should accept custom breakpoint options', () => {
         const customOptions = {
           namespace: 'v0:breakpoints',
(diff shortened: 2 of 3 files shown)


johnleider:
Thanks for this — you were first to fix #730, and our maintainer-side sweep opened a duplicate (#770, since merged) without catching your open PR. That overlap was our process miss, and I'm sorry your work sat while an equivalent landed. The merged fix reached the same behavior via a shared resolve/apply refactor and added a no-matchMedia fallback suite on top of the cases you covered, so I'm closing this as superseded by master rather than by preference.

PR #762, opened 2026-08-01
fix(useResizeObserver): report the content box on the immediate entry's contentRect

### LINKED ISSUE #729: [Bug] useResizeObserver: immediate entry reports the border box, every later entry reports the content box
### Problem

`useResizeObserver`'s `immediate` path synthesizes its first entry from `getBoundingClientRect()`, which reports the **border box in viewport coordinates**. Every entry the real observer delivers afterwards reports the **content box in padding-box coordinates**. A consumer reading `contentRect` therefore gets one set of numbers on mount and a different set on the first genuine resize, with no size change in between.

This contradicts the contract the file itself documents. `ResizeObserverEntry`'s JSDoc (`packages/0/src/composables/useResizeObserver/index.ts:50-55`) states:

> `contentRect` always describes the content box, regardless of the `box` option.

On the `immediate` path it does not.

### Evidence

All line numbers are `packages/0/src/composables/useResizeObserver/index.ts` at `adaace94d` (post-#725).

**Observer path** — `contentRect` copied straight from the native entry (`:227-239`):

```ts
create: cb => new ResizeObserver(entries => {
  cb(entries.map(e => ({
    contentRect: {
      width: e.contentRect.width,
      height: e.contentRect.height,
      top: e.contentRect.top,
      left: e.contentRect.left,
    },
    ...
```

**Immediate path** — `contentRect` built from a client rect (`:241` → `measure()` at `:125-167`):

```ts
immediate: options.immediate ? el => [measure(el)] : undefined,
```

```ts
function measure (el: Element): ResizeObserverEntry {
  const rect = el.getBoundingClientRect()
  ...
  return {
    contentRect: {
      width: rect.width,      // :158
      height: rect.height,
      top: rect.top,
      left: rect.left,
    },
    borderBoxSize: [{ inlineSize: inline.border, blockSize: block.border }],
    contentBoxSize: [{ inlineSize: inline.content, blockSize: block.content }],
    target: el,
  }
}
```

`measure()`'s own JSDoc (`:116-118`) records the decision: *"`contentRect` keeps its existing client-rect source so callers reading it see no change."* #725 deliberately left this alone to stay purely additive — the mismatch is pre-existing, not introduced by that PR.

### Three distinct discrepancies

| | `immediate` entry | Observer entry |
|---|---|---|
| Box model | border box (content + padding + border) | content box |
| `top` / `left` | viewport offsets | padding-box offsets (per spec, `padding-top` / `padding-left`) |
| CSS transforms | scaled by them (`getBoundingClientRect`) | unscaled layout values |

### Reproduction

```html
<div id="box" style="width: 30px; padding: 5px; box-sizing: content-box"></div>
```

```ts
const { width } = useElementSize(box)   // immediate: true is hardcoded (:292)

// on mount:            width.value === 40   (border box)
// after first resize:  width.value === 30   (content box)
// …with no actual size change in between
```

### Affected call sites

Only consumers that pass `immediate: true` **and** read the entry are affected. Inside the package that is the two `useElementSize` call sites — `useElementSize` hardcodes `{ immediate: true }` (`:292`) and reads `contentRect.width/height` (`:290-291`):

- `packages/0/src/composables/createOverflow/index.ts:170` — `useElementSize(container)`; the container width seeds `capacity` and `isOverflowing`, so an overflow menu can compute a different capacity on mount than on first resize. Downstream: `Overflow`, `AvatarGroup`.
- `packages/0/src/components/Carousel/CarouselViewport.vue:85` — `useElementSize(el)`; `width`/`height` feed `slideStep`, so the initial programmatic scroll step can be off by the viewport's padding + border.

Plus every external caller of `useElementSize`, and any external `useResizeObserver(el, cb, { immediate: true })` that reads `contentRect`.

**Not** affected, for the record — these were checked and either pass no `immediate` option or ignore the entry:

- `composables/createVirtual/index.ts:432` — no `immediate`
- `composables/useDragDrop/index.ts:561` — no `immediate`, callback ignores the entry
- `components/Splitter/SplitterPanel.vue:99` — no `immediate`
- `components/Avatar/AvatarRoot.vue:119`, `components/Avatar/AvatarIndicator.vue:88`, `components/Overflow/OverflowIndicator.vue:86` — no `immediate`, callbacks ignore the entry
- `components/Carousel/CarouselViewport.vue:111` — passes `immediate: true`, but `syncScrollPadding` reads `getComputedStyle` and never touches the entry

### Fix sketch

`measure()` already computes the correct content-box values from `getComputedStyle` for `contentBoxSize` (`:141-148`, `:164`). Sourcing `contentRect.width/height` from `width.content` / `height.content` closes the size gap; matching `top`/`left` to the padding offsets closes the coordinate gap. #725's `borderBoxSize` is now the supported way to ask for the old numbers.

### This is a behavior change

Fixing it changes the values the affected call sites above see on mount — on a padded or bordered element they get smaller. That is the point (they currently disagree with every subsequent entry), but it is a visible delta for anyone who compensated for the old numbers, and the release note should say so.

sridhar-3009:
Closes #729.

The `immediate` option synthesized its first entry from `getBoundingClientRect()`, which reports the **border box** in viewport coordinates. Every entry the real observer delivers afterwards reports the **content box**, positioned relative to the padding edge. This contradicts `contentRect`'s own documented contract:

> `contentRect` always describes the content box, regardless of the `box` option.

A consumer reading `contentRect` on mount (e.g. `useElementSize`, which hardcodes `immediate: true` and is used by `createOverflow` and `CarouselViewport`) could see one set of numbers on mount and a different set on the first real resize, with no actual size change in between.

## Fix

`measure()` already computes correct content-box values from `getComputedStyle` for `contentBoxSize`/`borderBoxSize`. `contentRect` now sources its `width`/`height` from the same content-box values, and its `top`/`left` from the padding offsets, matching what the observer's own entries report.

## Tests

- Extended the existing 'immediate entry matching what the observer reports' test to assert `contentRect` equality, not just the box-size arrays.
- Added a dedicated test asserting the immediate `contentRect` reports content-box dimensions positioned at the padding offset.

All 9 browser tests + 26 non-browser tests for this composable pass.

### DIFF
--- a/.changeset/fix-resize-observer-immediate-content-box-729.md
+++ b/.changeset/fix-resize-observer-immediate-content-box-729.md
@@ -1,0 +1,17 @@
+---
+"@vuetify/v0": patch
+---
+
+fix(useResizeObserver): report the content box on the immediate entry's contentRect (#729)
+
+The `immediate` option synthesized its first entry from `getBoundingClientRect()`,
+which reports the border box in viewport coordinates. Every entry the real observer
+delivers afterwards reports the content box, positioned relative to the padding edge —
+contradicting `contentRect`'s own documented contract ("always describes the content
+box, regardless of the `box` option"). A consumer reading `contentRect` on mount (for
+example `useElementSize`, which hardcodes `immediate: true`) could see one set of
+numbers on mount and a different set on the first real resize, with no size change in
+between.
+
+`measure()` now derives `contentRect` from the same computed-style values it already
+uses for `contentBoxSize`/`borderBoxSize`, so the immediate and observed paths agree.
--- a/packages/0/src/composables/useResizeObserver/index.browser.test.ts
+++ b/packages/0/src/composables/useResizeObserver/index.browser.test.ts
@@ -155,6 +155,20 @@ describe('useResizeObserver box model', () => {
 
     expect(size(immediate.borderBoxSize[0])).toEqual(size(observed.borderBoxSize[0]))
     expect(size(immediate.contentBoxSize[0])).toEqual(size(observed.contentBoxSize[0]))
+    expect(immediate.contentRect).toEqual(observed.contentRect)
+  })
+
+  it('should report the immediate contentRect on the content box, positioned at the padding offset', async () => {
+    const el = element()
+    const { next } = observe(el, { immediate: true })
+
+    const immediate = await next()
+
+    expect(immediate.contentRect.width).toBe(CONTENT.inlineSize)
+    expect(immediate.contentRect.height).toBe(CONTENT.blockSize)
+    // 16px padding on the inline edges, 4px on the block edges (see GEOMETRY)
+    expect(immediate.contentRect.left).toBe(16)
+    expect(immediate.contentRect.top).toBe(4)
   })
 
   it('should swap the logical axes on the immediate entry under a vertical writing mode', async () => {
--- a/packages/0/src/composables/useResizeObserver/index.ts
+++ b/packages/0/src/composables/useResizeObserver/index.ts
@@ -114,8 +114,9 @@ function pxToNumber (value: string | undefined, fallback = 0): number {
  * The box sizes come from `getComputedStyle` rather than
  * `getBoundingClientRect` because resolved lengths are layout values — like the
  * observer, and unlike a client rect, they are not scaled by CSS transforms.
- * `contentRect` keeps its existing client-rect source so callers reading it see
- * no change.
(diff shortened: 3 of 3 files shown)


johnleider:
Thanks — first fix filed for #729, and functionally identical to what merged (#769); our sweep missed your open PR and duplicated it, which is on us. Master carries your exact immediate-equals-observed assertion plus pinned literal expectations and a happy-dom synthesis test, so there's nothing here master lacks. Closing as superseded, with apologies for the overlap.

PR #763, opened 2026-08-01
fix(NumberField): default the spinbutton's accessible name

### LINKED ISSUE #741: [Bug] NumberField: spinbutton input has no accessible name in its documented shape
Found by the advisory axe-core sweep in #736. Rule: [`label`](https://dequeuniversity.com/rules/axe/4.12/label), **critical**, 1 element.

## What axe reports

```
label (critical) — 1 element in NumberField
    Form elements must have labels
    > <input id="v-0" role="spinbutton" type="text" inputmode="decimal" autocomplete="off" autocorrect="off" spellcheck="false" aria-valuenow="0" aria-valuetext="0" ...>
```

Against the shape `NumberField`'s own `@example` documents.

## Why this is not already covered by #640

PR #640 adds a `labelledBy` prop so consumers can point the spinbutton at an existing element via `aria-labelledby`. That closes the case where a label element exists elsewhere in the DOM.

It does not close this one. The `label` rule fires on a form element with **no** name from any source — no `aria-label`, no `aria-labelledby`, no wrapping or `for`-associated `<label>`, no title. The documented example supplies none of them, so the input ships nameless even after #640 lands, because #640 gives consumers a mechanism they still have to reach for.

Worth confirming the interaction before fixing: if #640 merges first, re-run the sweep to see whether this becomes purely an example/docs fix or still needs a component-side default.

## Options

Same fork as #740 (Slider) and #638 (Progress), and it should be answered the same way across all of them:

1. A component-side default name via `useLocale()` — `locale.ti('NumberField.label') ?? 'Number'` — matching the precedent already in `Pagination.Root`, `Breadcrumbs.Root`, `NumberField.Increment` and `NumberField.Decrement` (the increment/decrement buttons are already named this way; only the input itself is not, which makes the omission look accidental).
2. A `label` prop plus corrected examples, on the argument that a generic name is worse than a consumer-supplied one.

Note the asymmetry in favour of a fix: `NumberFieldIncrement.vue` already does `'aria-label': locale.ti('NumberField.increment') ?? 'Increment'`. The buttons around the input are named; the input is not.

## Related

- #640 — `labelledBy` prop, in flight, adjacent but not sufficient
- #740, #739, #638 — same class, other components

sridhar-3009:
Closes #741.

`NumberField`'s `role="spinbutton"` input reads `aria-label` from `root.label`, but that prop has no default — and `aria-labelledby` requires the consumer to point at an existing element. The shape `NumberField`'s own `@example` documents supplies neither, so the spinbutton ships with no accessible name at all (axe `label`, critical).

This is not covered by #640 (`labelledBy` prop, merged) — that closes the case where a label element exists elsewhere in the DOM, but does nothing when the consumer supplies no naming mechanism at all, which is exactly the documented default shape.

## Fix

`NumberField.Control` now falls back to `locale.ti('NumberField.label') ?? 'Number'` when neither `label` nor `ariaLabelledby` is set — matching the precedent already used by the increment/decrement buttons (`NumberField.increment` / `NumberField.decrement`), which are already named this way. A consumer-supplied `label` still wins, and `ariaLabelledby` still suppresses `aria-label` entirely.

## Tests

- Replaced the (now-incorrect) 'should not set aria-label when no label prop is provided' test with one asserting the default fallback.
- Added a test confirming an explicit `label` prop still wins over the default.

139 existing + new component tests pass.

### DIFF
--- a/.changeset/fix-numberfield-spinbutton-accessible-name-741.md
+++ b/.changeset/fix-numberfield-spinbutton-accessible-name-741.md
@@ -1,0 +1,16 @@
+---
+"@vuetify/v0": patch
+---
+
+fix(NumberField): default the spinbutton's accessible name (#741)
+
+`NumberField`'s `label` and `ariaLabelledby` props both collapsed to `undefined`
+when a consumer supplied neither — including the shape `NumberField`'s own
+`@example` documents — so the `role="spinbutton"` input shipped with no
+accessible name at all (axe `label`, critical).
+
+`NumberField.Control` now falls back to a locale-driven default (`locale.ti('NumberField.label') ?? 'Number'`)
+when no `label` or `ariaLabelledby` is provided, matching the precedent already
+set by the increment/decrement buttons (`NumberField.increment`/`NumberField.decrement`).
+A consumer-supplied `label` still takes priority, and `ariaLabelledby` still
+suppresses `aria-label` entirely.
--- a/packages/0/src/components/NumberField/NumberFieldControl.vue
+++ b/packages/0/src/components/NumberField/NumberFieldControl.vue
@@ -15,6 +15,9 @@
   // Context
   import { useNumberFieldRoot } from './NumberFieldRoot.vue'
 
+  // Composables
+  import { useLocale } from '#v0/composables/useLocale'
+
   // Utilities
   import { isNull } from '#v0/utilities'
   import { mergeProps, onMounted, shallowRef, toRef, useAttrs, watch } from 'vue'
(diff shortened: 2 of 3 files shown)


johnleider:
Thanks — the implementation line here is byte-identical to what merged (#772), and you filed first; the duplication was our sweep's miss. The merged version also added the NumberField.label key to the en locale catalog so the default name is translatable rather than hardcoded-fallback-only. Closing as superseded by master, with apologies for the overlap.

PR #764, opened 2026-08-01
fix(Slider): default the thumb's accessible name

### LINKED ISSUE #740: [Bug] Slider: role="slider" thumb ships with no accessible name by default
Found by the advisory axe-core sweep in #736. Rule: [`aria-input-field-name`](https://dequeuniversity.com/rules/axe/4.12/aria-input-field-name), **serious**, 1 element.

## What axe reports

```
aria-input-field-name (serious) — 1 element in Slider
    ARIA input fields must have an accessible name
    > <span role="slider" tabindex="0" aria-valuenow="50" aria-valuemin="0" aria-valuemax="100" aria-orientation="horizontal" aria-disabled="false" data-state="idle">
```

Against the shape `Slider`'s own `@example` documents:

```vue
<Slider.Root v-model="value" :min="0" :max="100">
  <Slider.Track><Slider.Range /></Slider.Track>
  <Slider.Thumb />
</Slider.Root>
```

## Why

`SliderRoot.vue` already has the props:

```ts
'aria-label': label || undefined,
'aria-labelledby': ariaLabelledby || undefined,
```

Both collapse to `undefined` when the consumer passes neither, and neither has a default. `aria-valuetext` is emitted but is the value, not the name, so the thumb ships with `role="slider"` and nothing to announce it by.

This is not the same defect as the Rating naming gap: Slider *has* the mechanism and no fallback. Rating has no mechanism at all.

## Options

1. **Locale-driven default**, matching how the rest of the library names its widgets — `locale.ti('Slider.label') ?? 'Slider'`, the same shape `Pagination.Root`, `Breadcrumbs.Root` and `Carousel.Item` already use. Guarantees a name, at the cost of a generic one.
2. **Document it and leave it**, on the grounds that a headless library should not invent a name for a control whose meaning only the consumer knows. Then the `@example` and the docs page must both show `label`, since the documented shape is what consumers copy — and what this sweep audits.

Option 1 matches existing precedent in the codebase and is the low-friction fix; option 2 is defensible but only if the examples change, otherwise the violation just moves out of CI and into every consumer's app.

Note this is the same trade-off `Progress.Root` resolves in #638 by adding an `ariaLabel` prop — worth landing the same answer across all three (Progress, Slider, Rating) rather than three different ones.

## Related

- #739 — Rating, same rule, no naming API at all
- #638 — Progress, same class of defect, fix in flight

sridhar-3009:
Closes #740.

`SliderRoot.vue` already has `label`/`ariaLabelledby` props, but they land on the `role="group"` wrapper — the `role="slider"` thumb itself reads its name only from its own `ariaLabel`/`ariaLabelledby` props, which default to `undefined`. The shape `Slider`'s own `@example` documents (`<Slider.Thumb />` with no `ariaLabel`) therefore ships a thumb with no accessible name at all (axe `aria-input-field-name`, serious). `aria-valuetext` is populated but is the value, not the name.

## Fix

`Slider.Thumb` now falls back to `locale.ti('Slider.label') ?? 'Slider'` when neither `ariaLabel` nor `ariaLabelledby` is set — the same locale-driven-default pattern already used by `Pagination.Root`, `Breadcrumbs.Root`, and `Carousel.Item`. A consumer-supplied `ariaLabel` still wins. Also tightened `aria-labelledby`/`aria-label` to be mutually exclusive (previously both could be emitted at once when a consumer passed both props).

## Tests

- Added a test asserting the default `aria-label` fallback.
- Added a test asserting `ariaLabelledby` suppresses the default `aria-label`.

92 existing + new browser tests pass.

### DIFF
--- a/.changeset/fix-slider-thumb-accessible-name-740.md
+++ b/.changeset/fix-slider-thumb-accessible-name-740.md
@@ -1,0 +1,16 @@
+---
+"@vuetify/v0": patch
+---
+
+fix(Slider): default the thumb's accessible name (#740)
+
+`Slider.Thumb` renders `role="slider"` and read its accessible name only from
+the consumer-supplied `ariaLabel`/`ariaLabelledby` props, with no fallback —
+so the shape `Slider`'s own `@example` documents shipped a thumb with no name
+at all (axe `aria-input-field-name`, serious).
+
+`Slider.Thumb` now falls back to `locale.ti('Slider.label') ?? 'Slider'` when
+neither `ariaLabel` nor `ariaLabelledby` is provided, matching the pattern
+already used by `Pagination.Root`, `Breadcrumbs.Root` and `Carousel.Item`. A
+consumer-supplied `ariaLabel` still wins, and `ariaLabelledby` still suppresses
+`aria-label` entirely (previously both attributes could be set at once).
--- a/packages/0/src/components/Slider/SliderThumb.vue
+++ b/packages/0/src/components/Slider/SliderThumb.vue
@@ -16,6 +16,9 @@
   // Context
   import { useSliderRoot } from './SliderRoot.vue'
 
+  // Composables
+  import { useLocale } from '#v0/composables/useLocale'
+
   // Utilities
   import { isUndefined } from '#v0/utilities'
   import { mergeProps, onBeforeUnmount, toRef, toValue, useAttrs, useTemplateRef } from 'vue'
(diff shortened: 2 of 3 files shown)


johnleider:
Thanks — you were first on #740, and our sweep duplicated it (#771, merged) without catching your PR; that's our miss. Your version also got one thing more right than ours: when both ariaLabel and ariaLabelledby are set, your ordering suppressed aria-label so labelledby is the sole name — we merged the opposite ordering and shipped an inconsistency with NumberField. #784 unifies all three components on your ordering and credits this PR for surfacing it. Closing as superseded by master + #784.

PR #765, opened 2026-08-01
fix(Select, Combobox): correct the documented Item example to avoid double role=option

### LINKED ISSUE #738: [Bug] Select.Item renders role="option" twice — nested options have no required listbox parent
Found by the advisory axe-core sweep in #736. Rule: [`aria-required-parent`](https://dequeuniversity.com/rules/axe/4.12/aria-required-parent), **critical**, 3 elements.

## What axe reports

```
aria-required-parent (critical) — 3 elements in Select
    Certain ARIA roles must be contained by particular parents
    > <div id="v-0-option-apple" role="option" aria-selected="false" aria-disabled="false" data-id="apple">Apple</div>
    > <div id="v-0-option-banana" role="option" aria-selected="false" aria-disabled="false" data-id="banana">Banana</div>
    > <div id="v-0-option-cherry" role="option" aria-selected="false" aria-disabled="false" data-id="cherry">Cherry</div>
```

## Why

`Select.Content` correctly renders `role="listbox"`, and the items are nested inside it. The failure is one level lower: **every option is rendered twice.**

`SelectItem.vue` binds its whole `attrs` bundle onto its own `Atom`, and `as` defaults to `'div'` with `renderless` unset:

```vue
<!-- packages/0/src/components/Select/SelectItem.vue -->
<template>
  <Atom :as :renderless v-bind="slotProps.attrs">
    <slot v-bind="slotProps" />
  </Atom>
</template>
```

`Select.Item`'s own `@example` in the barrel then spreads that same `attrs` object onto a child:

```vue
<!-- packages/0/src/components/Select/index.ts -->
<Select.Item id="apple" value="Apple" v-slot="{ isSelected, attrs }">
  <div v-bind="attrs">Apple {{ isSelected ? '✓' : '' }}</div>
</Select.Item>
```

Result: `listbox > div[role="option"] > div[role="option"]`. The inner option's nearest ancestor with a role is another `option`, not the `listbox`, so it has no required parent — the three elements axe names above are the inner divs.

## This is also the double-fire hazard, already documented

`.claude/rules/components.md` states the rule this example breaks:

> Consumers must **only** spread slot `attrs` onto their own element when using `renderless` mode. In non-renderless mode, spreading `attrs` onto a child element causes handlers to fire twice — once on the child, then again on the Atom wrapper via event bubbling.
>
> When writing examples for new components, never `v-bind="attrs"` on children inside a non-renderless component.

So `Select.Item`'s documented usage ships a duplicated `onClick` alongside the duplicated `role`. One fix addresses both.

## Suggested fix

Correct the example rather than the component — either drop the inner wrapper, or mark the Item `renderless` so the consumer's element is the only one rendered:

```vue
<Select.Item id="apple" value="Apple" renderless v-slot="{ isSelected, attrs }">
  <div v-bind="attrs">Apple {{ isSelected ? '✓' : '' }}</div>
</Select.Item>
```

Both the `@example` JSDoc in `packages/0/src/components/Select/index.ts` and `apps/docs/src/examples/components/select/` need the same correction, and the sweep fixture at `packages/0/src/components/fixtures/Select.vue` follows whichever shape lands.

## `Combobox.Item` is structurally identical and unverified

`ComboboxItem.vue` has the same `as = 'div'` default, the same `v-bind="slotProps.attrs"` on its Atom, and the same `<div v-bind="attrs">` in its documented example. It did **not** appear in the sweep's report, but that is a gap in the harness rather than evidence it is clean: the sweep opens a collapsed widget by clicking `[aria-expanded="false"]`, and Combobox's listbox does not open from that click, so its options were never in the DOM to audit. Check Combobox by hand when fixing this, per the bug-family rule.

## Not in scope

The `button-name` violations the same sweep reports on `Select.Activator` are the Select half of #613 (PR #635), and the accessible-name mechanism is #610. This issue is only the nested-`option` structure.

sridhar-3009:
Closes #738.

`Select.Item` (and, identically, `Combobox.Item`) binds its whole `attrs` bundle — including `role="option"` — onto its own `Atom`, which is not `renderless` by default. Both components' `@example` JSDoc then spread that same `attrs` object onto a child `<div>` inside the default slot:

```vue
<Select.Item id="apple" value="Apple" v-slot="{ isSelected, attrs }">
  <div v-bind="attrs">Apple {{ isSelected ? '✓' : '' }}</div>
</Select.Item>
```

`.claude/rules/components.md` documents this as unsupported outside `renderless` mode — spreading `attrs` onto a child in non-renderless mode renders the ARIA role twice (`listbox > div[role="option"] > div[role="option"]`, failing axe's `aria-required-parent`) and double-fires every handler in `attrs` once on the child and again via bubbling to the Atom. The documented shape is exactly this bug.

## Fix

Added `renderless` to the documented `Select.Item` / `Combobox.Item` usage in both components' `@example` blocks (the top-level module example and each component's own JSDoc), so the consumer's own element is the only one rendered. This matches how `packages/0/src/components/fixtures/Select.vue` / `Combobox.vue` and every real docs example (`apps/docs/src/examples/components/select/*`, `combobox/*`) already use these components — none of them spread `attrs` on a child, so none were actually affected by the bug; only the documented `@example` shape was.

Per the issue, `Combobox.Item` is structurally identical to `Select.Item` (same `as='div'` default, same non-renderless `attrs` spread) and didn't appear in the #736 sweep only because the harness never got Combobox's listbox open to audit — checked it by hand here and it has the identical example bug, now fixed alongside Select.

Documentation-only change; the fixture used by the a11y sweep was already correct, so all 53 existing `a11y.browser.test.ts` tests pass unchanged.

### DIFF
--- a/.changeset/fix-select-combobox-item-double-role-738.md
+++ b/.changeset/fix-select-combobox-item-double-role-738.md
@@ -1,0 +1,18 @@
+---
+"@vuetify/v0": patch
+---
+
+fix(Select, Combobox): correct the documented Item example to avoid double `role="option"` (#738)
+
+`Select.Item` and `Combobox.Item` already bind their full `attrs` bundle (including
+`role="option"`) onto their own non-renderless `Atom`. Both components' `@example`
+JSDoc then spread that same `attrs` object onto a child `<div>` inside the default
+slot — which `.claude/rules/components.md` documents as unsupported outside
+`renderless` mode: it renders `role="option"` twice (`listbox > div[role="option"] >
+div[role="option"]`, failing axe's `aria-required-parent`) and double-fires every
+handler in `attrs`, once on the child and again via bubbling to the Atom.
+
+Added `renderless` to the documented `Select.Item` / `Combobox.Item` usage in both
+components' `@example` blocks, so the consumer's own element is the only one
+rendered — matching how `packages/0/src/components/fixtures/Select.vue` and
+`Combobox.vue` (and every real docs example) already use these components.
--- a/packages/0/src/components/Combobox/index.ts
+++ b/packages/0/src/components/Combobox/index.ts
@@ -61,6 +61,7 @@ import Root from './ComboboxRoot.vue'
  *         :key="item.id"
  *         :id="item.id"
  *         :value="item.label"
+ *         renderless
  *         v-slot="{ isSelected, attrs }"
  *       >
  *         <div v-bind="attrs">{{ item.label }}</div>
@@ -171,7 +172,7 @@ export const Combobox = {
    *
    * @example
    * ```vue
-   * <Combobox.Item id="apple" value="Apple" v-slot="{ isSelected, isHighlighted, attrs }">
+   * <Combobox.Item id="apple" value="Apple" renderless v-slot="{ isSelected, isHighlighted, attrs }">
    *   <div v-bind="attrs" :class="{ highlighted: isHighlighted }">
    *     Apple {{ isSelected ? '✓' : '' }}
    *   </div>
--- a/packages/0/src/components/Select/index.ts
+++ b/packages/0/src/components/Select/index.ts
@@ -57,6 +57,7 @@ import Value from './SelectValue.vue'
  *         :key="item.id"
  *         :id="item.id"
  *         :value="item.label"
+ *         renderless
  *         v-slot="{ isSelected, attrs }"
  *       >
  *         <div v-bind="attrs">{{ item.label }}</div>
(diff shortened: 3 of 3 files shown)


johnleider:
Thanks — right diagnosis on #738. The merged fix (#775) took the other repair (dropping the inner element rather than adding renderless, so the documented shape has one option element and one click binding) and swept the same defect family through Radio, Treeview, and the Combobox examples. Since master's examples no longer have the inner spread at all, closing as superseded — and again, you were first; the duplicate was our miss.

PR #768, opened 2026-08-03
fix(Input): make type prop reactive on InputRoot context

### LINKED ISSUE #757: InputRoot: type prop captured non-reactively — post-mount type changes are silently inert
## Bug

`Input.Root`'s `type` prop is captured as a plain, non-reactive field on `InputRootContext`: the context interface declares `readonly type: string` (packages/0/src/components/Input/InputRoot.vue — every sibling field is a ref/getter), and setup assigns the destructured prop value once into the provided context object. `Input.Control` reads that frozen value, so changing `:type` on `Input.Root` after mount never reaches the rendered control element.

## Reproduction

Bind `:type` to a ref, toggle `'password'` ↔ `'text'` (the classic reveal-password toggle), and inspect the `<input>`'s DOM `type` attribute — it never changes. Verified via `document.querySelector('input').type` before/after toggle: icon/state around it updates, the control does not.

## Expected

`type` should be reactive like the context's other fields (wrap with `toRef` when placed on the context, and have `InputControl` read it reactively).

## Context

Found while building a downstream design-system field component with a reveal toggle; worked around with `<Input.Root :key="type">` (full remount — costs transient validation state + focus on toggle), so not urgent downstream, but the primitive's contract implies reactivity.

johnleider:
`Input.Root` captured the `type` prop once at setup as a plain `readonly type: string` on `InputRootContext` — the only non-reactive field on the context. Binding `:type` to a ref and toggling it after mount (e.g. `'password'` ↔ `'text'`) never updated the rendered `<input>`'s DOM type.

- `InputRootContext.type` is now `Readonly<Ref<string>>`, provided via `toRef(() => type)` like the sibling fields
- `Input.Control` reads it reactively, so the control's `type` attribute tracks post-mount changes
- Regression test added to the colocated browser suite; changeset included (patch)

Closes #757

### DIFF
--- a/.changeset/fix-input-root-type-reactivity.md
+++ b/.changeset/fix-input-root-type-reactivity.md
@@ -1,0 +1,7 @@
+---
+"@vuetify/v0": patch
+---
+
+fix(Input): make the `type` prop reactive (#757)
+
+`Input.Root`'s `type` prop was captured once at setup, so binding `:type` to a ref and toggling it after mount (e.g. `'password'` ↔ `'text'`) never updated the rendered `<input>`'s type. The prop now flows reactively through the root context, and `Input.Control` reflects post-mount changes.
--- a/packages/0/src/components/Input/InputControl.vue
+++ b/packages/0/src/components/Input/InputControl.vue
@@ -81,7 +81,7 @@
 
     return {
       'id': root.id,
-      'type': root.type,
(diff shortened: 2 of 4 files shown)


johnleider:
Superseded by #766 — identical fix, filed first, with broader tests.

PR #787, opened 2026-08-04
feat(Otp): add Otp compound component over createOtp

### LINKED ISSUE #519: [Feature Request] Otp component (over createOtp)
Build the **Otp** compound component over the shipped `createOtp` composable (preview, since 1.0.0-alpha.5).

Part of the **v1.3.0** epic (#509).

## Checklist — [new-feature checklist](.claude/rules/new-feature-checklist.md), components
- [ ] `packages/0/src/components/Otp/` — `.vue` files + `index.ts` barrel
- [ ] `packages/0/src/components/index.ts` — add `export * from './Otp'` (alphabetical)
- [ ] `packages/0/src/maturity.json` — flip `Otp` `draft` → `preview`, `since: null`, add `description`
- [ ] `apps/docs/src/pages/components/forms/otp.md` — docs page with frontmatter
- [ ] `apps/docs/src/examples/components/otp/basic.vue` — basic example
- [ ] `apps/docs/src/pages/components/index.md` — table entry (forms section)
- [ ] `packages/0/README.md` — What's Included table
- [ ] root `README.md` — keep in sync
- [ ] `packages/0/src/surface.test.ts` — add exports to `COMPONENTS` freeze array (sorted)
- [ ] `<DocsApi />` renders + feature appears in `api-whitelist.ts` after build
- [ ] `pnpm build:0 && pnpm typecheck && pnpm lint:fix && pnpm build:docs` green

sridhar-3009:
Closes #519.

`createOtp` has shipped a complete, tested headless primitive for one-time-password / verification-code values since 1.0.0-alpha.5 (pattern-gated entry, completion detection, decisional async `onComplete`), but there was no Vue component wrapping it — consumers had to hand-roll rendering, focus management, and keyboard/paste wiring themselves.

## What's added

`Otp.Root`, `Otp.Item`, `Otp.HiddenInput` — same Root/Item/HiddenInput shape as `Rating` and `Slider`:

- **`Otp.Root`** creates the `createOtp` context, bridges v-model, renders a `role="group"` container with a locale-driven default `aria-label` ("Verification code"), and tracks each Item's element so focus can move between boxes.
- **`Otp.Item`** renders one character box (`<input maxlength="1" autocomplete="one-time-code">` by default — `autocomplete="one-time-code"` is the documented convention that lets mobile OSes offer SMS-code autofill). Handles:
  - auto-advance to the next box when a pattern-accepted character is typed
  - Backspace on an empty box moves focus back and clears the previous box (native Backspace on a *filled* box just clears it in place, standard OTP UX)
  - ArrowLeft/ArrowRight move focus directly between boxes
  - paste distributes the clipboard text across boxes starting at the pasted-into box (via `createOtp`'s `distribute`), since `maxlength` would otherwise silently truncate a multi-character paste to 1 char
  - each box gets its own `aria-label` ("Digit N of length")
- **`Otp.HiddenInput`** mirrors `Rating.HiddenInput` — auto-rendered when `name` is set, for native form submission.

Registered in the advisory a11y sweep (`packages/0/src/components/fixtures/Otp.vue`) — clean, no axe violations. Added the two new locale keys (`Otp.label`, `Otp.itemLabel`) to the en catalog per the static locale-coverage check. Bumped `Otp`'s `maturity.json` entry from `draft` to `preview`.

## Tests

23 tests in `Otp/index.test.ts` covering rendering, v-model (including auto-advance, backspace-back, arrow-key nav, paste distribution, and the beforeinput pattern gate), disabled/readonly no-ops, the `name`/hidden-input path, and `onComplete` (both the completion signal and async rejection). 94%/98.5% statement/line coverage on the new files, consistent with other components in this codebase.

Full suite: 4819 non-browser tests pass, 54 a11y-sweep browser tests pass, typecheck/knip/sherif clean.


packages/0/src/components/Otp/OtpItem.vue
```diff
@@ -0,0 +1,209 @@
...
+  function onPaste (e: ClipboardEvent) {
+    if (root.isDisabled.value || root.isReadonly.value) return
+    e.preventDefault()
+    const text = e.clipboardData?.getData('text') ?? ''
+    const written = root.distribute(text, index)
+    // distribute splices at min(value.length, index), so the next empty box
+    // sits at that start plus the written count — `index + written` would
+    // overshoot when the paste lands before `index` on a shorter value.
+    // Computed from the pre-write value: the model ref may update async.
+    if (written > 0) root.focusItem(Math.min(root.value.value.length, index) + written)
```
johnleider: **[bug]** Paste focus is computed from the **post-write** value, so it overshoots whenever `distribute` lands before `index` (empty/short value, paste into a later box). `createOtp.distribute` assigns `value.value = next` synchronously; Vue refs do not update async. For value `''`, paste `'12'` at index 3: written=2, post-write length=2, `Math.min(2, 3)+2 = 4`, so focus moves to box 5 instead of the next empty box (index 2). The test at `index.test.ts:303-313` documents the intended focus (`itemEls()[2]`) and will fail against this formula. The comment on 168-171 describes the correct pre-write intent.

**Suggestion:** Capture length before `distribute`, or just `focusItem(root.value.value.length)` after it (the next empty index is the new joined length). `focusItem` already clamps at `length-1`.

packages/0/src/components/Otp/OtpItem.vue
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    const text = target.value
+
+    if (text === '') {
+      root.write(index, '')
+      return
+    }
+
+    const entered = text.at(-1)!
```
johnleider, line 151: **[bug]** Multi-character `input` (Chrome/Safari SMS autofill, `insertReplacementText`, some mobile IMEs) is collapsed to `text.at(-1)` and written as a single cell. `onBeforeinput` (line 106) explicitly lets `e.data.length !== 1` through, then this path keeps only the last character — so an autofilled `"123456"` becomes `"6"`. Paste is handled; autofill usually does not fire `paste`. This is the primary OTP input path.

**Suggestion:** If `text.length > 1`, call `root.distribute(text, index)` (or `root.fill(text)` when replacing the whole code) and focus past the last written cell. Keep the single-char `write` path for real keystrokes.

packages/0/src/components/Otp/OtpItem.vue
```diff
@@ -0,0 +1,209 @@
...
+  const slotProps = toRef((): OtpItemSlotProps => ({
+    value: char.value,
+    state: state.value,
+    isDisabled: root.isDisabled.value,
+    isReadonly: root.isReadonly.value,
+    attrs: {
+      'type': 'text',
+      'inputmode': root.pattern.value === 'numeric' ? 'numeric' : 'text',
+      'autocomplete': 'one-time-code',
+      'maxlength': 1,
```
johnleider, line 214: **[bug]** Boxes are `maxlength: 1` with no select-on-focus and no overwrite in `beforeinput`. On desktop Chrome, focusing a filled cell leaves the caret at the end; the next keystroke is blocked by maxlength, `onInput` never fires, and the user cannot replace the digit without Backspace first. That is the normal correction gesture for an OTP field.

**Suggestion:** Select the box contents on focus (`(e.target as HTMLInputElement).select()`), and/or on `beforeinput` with a single accepted character `preventDefault()`, `write(index, e.data)`, `focusItem(index + 1)` so overwrite works even if the UA does not replace.

packages/0/src/maturity.json
```diff
@@ -561,8 +561,10 @@
       "description": "Numeric input with increment/decrement buttons, drag-to-scrub, and locale-aware formatting via `Intl.NumberFormat` for currency, percent, and units."
     },
     "Otp": {
-      "level": "draft",
-      "category": "forms"
+      "level": "preview",
```
johnleider, line 570: **[suggestion]** Promoted `Otp` draft → preview (`since: null` is correct) without the rest of the new-feature checklist. `draft → preview` requires a docs page and a basic example. Missing: `apps/docs/src/pages/components/forms/otp.md`, `apps/docs/src/examples/components/otp/basic.vue`, a Forms row in `apps/docs/src/pages/components/index.md`, and What's Included rows in both `packages/0/README.md` and root `README.md`. Barrel JSDoc already `@see`s `https://0.vuetifyjs.com/components/forms/otp` (404).

**Suggestion:** Add the docs page (Usage + Anatomy + Examples + Accessibility + FAQ + `<DocsApi />`), a `basic.vue` example, the index/README table rows, and keep `preview` / `since: null`.

packages/0/src/components/Otp/OtpRoot.vue
```diff
@@ -0,0 +1,212 @@
...
+    namespace?: string
+  }
+
+  export interface OtpRootSlotProps {
+    /** The joined OTP value */
+    value: string
+    /** Number of characters */
+    length: number
+    /** Whether the value is complete and pattern-valid */
+    isComplete: boolean
```
johnleider, line 107: **[suggestion]** `onComplete` resolving `false` is a featured API (clears value, sets `input` error `'Invalid code'`), but slot props only expose `isComplete` / `data-complete`. After a reject the group looks like a blank field — no `errors`, `isError`, `data-error`, or `aria-invalid`. Consumers cannot style or announce the failure without reaching through `useOtpRoot().input`.

**Suggestion:** Surface `otp.input.errors` / invalid state on Root slot props and attrs (`data-error`, `aria-invalid`), matching Input/NumberField.

packages/0/src/components/Otp/index.test.ts
```diff
@@ -0,0 +1,707 @@
+import { describe, expect, it, vi } from 'vitest'
```
johnleider: **[suggestion]** Component tests live in `index.test.ts` (happy-dom `v0:unit`), not `index.browser.test.ts`. Rating/NumberField do the same, and the a11y fixture covers axe — so this is not a merge blocker — but this suite is almost entirely focus, paste, and caret behavior, which is exactly what `v0:browser` exists for. `mountOtp` also `attachTo: document.body` (line 38) with no `wrapper.unmount()` / `afterEach`, so focused inputs leak across cases.

**Suggestion:** Move (or add) the focus/paste/autofill cases to `index.browser.test.ts`; unmount in `afterEach`. Keep the a11y fixture.

.changeset/feat-otp-component-519.md
```diff
@@ -0,0 +1,30 @@
...
+  by default). Typing a pattern-accepted character auto-advances focus to the next box;
+  Backspace on an empty box moves focus back and clears the previous box; arrow keys move
+  focus directly between boxes; pasting distributes the clipboard text across boxes
+  starting at the box it was pasted into (`createOtp.distribute`), matching what
+  `maxlength` would otherwise silently truncate. Each box gets its own
+  `aria-label` ("Digit N of length").
+- `Otp.HiddenInput` mirrors `Rating.HiddenInput` — auto-rendered when `name` is set, for
+  native form submission.
+
+Registered in the advisory a11y sweep (`packages/0/src/components/fixtures/Otp.vue`) —
```
johnleider: **[suggestion]** Changeset body includes mechanism that the releasing skill says must not land in the changelog: sibling-shape mirroring, the a11y sweep path, and the `maturity.json` draft→preview bump. Title `feat(Otp): add Otp compound component over createOtp (#519)` is fine.

**Suggestion:** Keep the consumer-facing behavior (Root/Item/HiddenInput, auto-advance, paste distribute, `onComplete`, `name` → hidden input). Drop the maturity bump, fixture path, and "same shape as Rating/Slider" lines.

johnleider (approved):
Follow-up commits addressed the original findings. Approving.

PR #788, opened 2026-08-04
feat(Virtualizer): add Virtualizer compound component over createVirtual

### LINKED ISSUE #517: [Feature Request] Virtualizer component (over createVirtual)
Build the **Virtualizer** compound component over the shipped `createVirtual` composable (preview, since 0.1.0).

Part of the **v1.3.0** epic (#509).

## Checklist — [new-feature checklist](.claude/rules/new-feature-checklist.md), components
- [ ] `packages/0/src/components/Virtualizer/` — `.vue` files + `index.ts` barrel
- [ ] `packages/0/src/components/index.ts` — add `export * from './Virtualizer'` (alphabetical)
- [ ] `packages/0/src/maturity.json` — flip `Virtualizer` `draft` → `preview`, `since: null`, add `description`
- [ ] `apps/docs/src/pages/components/data/virtualizer.md` — docs page with frontmatter
- [ ] `apps/docs/src/examples/components/virtualizer/basic.vue` — basic example
- [ ] `apps/docs/src/pages/components/index.md` — table entry (data section)
- [ ] `packages/0/README.md` — What's Included table
- [ ] root `README.md` — keep in sync
- [ ] `packages/0/src/surface.test.ts` — add exports to `COMPONENTS` freeze array (sorted)
- [ ] `<DocsApi />` renders + feature appears in `api-whitelist.ts` after build
- [ ] `pnpm build:0 && pnpm typecheck && pnpm lint:fix && pnpm build:docs` green

sridhar-3009:
Closes #517.

`createVirtual` has shipped a complete virtual-scrolling composable since 0.1.0 (visible-window rendering, fixed or dynamic item heights, bidirectional scrolling, scroll anchoring across data changes, edge detection for infinite scroll, iOS momentum/elastic scrolling), but there was no Vue component wrapping it — consumers had to wire up the scroll container, spacer elements, and resize reporting by hand, exactly as shown in the composable's own `@example`.

## What's added

`Virtualizer.Root` and `Virtualizer.Item`:

- **`Virtualizer.Root`** creates the `createVirtual` context and renders the scroll container plus the leading/trailing spacer elements that reserve space for items scrolled out of view. The default slot receives only the currently visible (+ overscan) items — iterate over them with `Virtualizer.Item`.
- **`Virtualizer.Item`** wraps a single rendered item and measures its own height via `ResizeObserver`, reporting it back to Root's `resize(index, height)` automatically — variable-height content just works without the consumer calling `resize()` by hand.

## Two things the a11y sweep caught during development (worth calling out)

1. **`scrollable-region-focusable`** — a scrollable container needs keyboard access. Root now defaults `tabindex="0"` on the container.
2. **`aria-allowed-attr`** — I initially had `Virtualizer.Item` auto-apply `aria-posinset`/`aria-setsize`, which axe correctly flagged: those attributes require a role that supports them (`option`/`row`/`gridcell`/etc.), and a bare element has none. A headless component can't know which role fits the consumer's content ahead of time, so I dropped that and rely on normal attrs passthrough instead — a consumer building a listbox or grid supplies `role`, `aria-posinset`, and `aria-setsize` directly on their `Virtualizer.Item` usage, which already reaches the rendered element through the existing `attrs` merge (covered by a test).

Registered in the advisory a11y sweep (`packages/0/src/components/fixtures/Virtualizer.vue`) — clean now. Bumped `Virtualizer`'s `maturity.json` entry from `draft` to `preview`.

## Tests

6 tests covering spacer rendering, visible-item slicing, passthrough ARIA attrs, resize reporting (isolated `Virtualizer.Item` test with a mocked root context), `scrollTo`/`reset` exposure, and reactive `items` changes. Full suite: 4802 non-browser tests pass, 54 a11y-sweep browser tests pass, typecheck/knip/sherif clean.


packages/0/src/components/Virtualizer/VirtualizerRoot.vue
```diff
@@ -0,0 +1,179 @@
...
+  })
+
+  const slotProps = toRef((): VirtualizerRootSlotProps => ({
+    items: virtual.items.value,
+    state: virtual.state.value,
+    scrollTo: virtual.scrollTo,
+    reset: virtual.reset,
+    attrs: {
+      tabindex: 0,
+      style: { overflowY: 'auto', height: height.value },
```
johnleider: **[bug]** `slotProps.attrs.style` always includes `height: height.value`. When the `height` prop is omitted that is `undefined`. `mergeProps(attrs, slotProps.attrs)` runs `normalizeStyle([consumerStyle, { overflowY: 'auto', height: undefined }])`, so a later `undefined` overwrites the consumer's height; Vue's `patchStyle` then does `setStyle(..., "")` and clears it. The public example (`Virtualizer.Root … style="height: 600px"`) is this path. A class-based height (`h-[600px]`) would survive; the style/prop path the barrel documents would not. CarouselViewport already avoids this by omitting optional style keys (`...(cond ? { key: val } : {})`).

**Suggestion:** Only put `height` on the style object when `_height` is set (`...(height.value ? { height: height.value } : {})`). Add a test that mounts with `style: { height: '600px' }` and no `height` prop and asserts the inline height remains `600px`.

packages/0/src/components/Virtualizer/VirtualizerRoot.vue
```diff
@@ -0,0 +1,179 @@
...
+</script>
+
+<template>
+  <Atom
+    ref="container"
+    v-bind="mergeProps(attrs, slotProps.attrs)"
+    :as
+    :renderless
+  >
+    <div data-spacer="start" :style="{ height: `${virtual.offset.value}px` }" />
```
johnleider: **[bug]** Leading/trailing spacer `<div>`s are hard-coded as Atom children. In renderless mode Atom renders its default slot with no wrapper (`Atom.vue` `v-if="renderless"`), so those spacers become fragment siblings of whatever the consumer returns from Root's slot — outside the element they bind `slotProps.attrs` onto. `createVirtual` also needs `offset`/`size` on that same scroll container, but slot props only expose `items`, `state`, `scrollTo`, `reset`, `attrs`. The renderless test only asserts `element` is cleared; it does not assert spacers still window the list. Image already warns when a mode that requires a wrapper is combined with `renderless`.

**Suggestion:** Either (a) skip spacer DOM when `renderless` and add `offset`/`size` to `VirtualizerRootSlotProps` so the consumer can place them inside their own scroller, or (b) treat renderless as unsupported here and warn like Image. The default (non-renderless) path can keep the spacers.

packages/0/src/components/Virtualizer/VirtualizerItem.vue
```diff
@@ -0,0 +1,81 @@
...
+    index,
+    namespace = 'v0:virtualizer:root',
+  } = defineProps<VirtualizerItemProps>()
+
+  const root = useVirtualizerRoot(namespace)
+
+  const itemRef = useTemplateRef<AtomExpose>('item')
+  const el = toRef(() => toElement(itemRef.value?.element) ?? null)
+
+  useResizeObserver(el, entries => {
```
johnleider: **[bug]** Item reports `entry.contentRect.height` (content box, excludes padding and border) as the row's layout height. Rows stack in normal flow; the space they occupy is the border box (`offsetHeight` / `borderBoxSize[0].blockSize`). Headless consumers will add padding/borders — the Item `@example` itself uses `class="p-2 border-b"`. Under UnoCSS preflight (`box-sizing: border-box`) that under-reports by padding+border per row, so spacers drift and the scrollbar/window desync. `useResizeObserver` documents this split and the `{ box: 'border-box' }` + `borderBoxSize` path.

**Suggestion:** Observe with `{ box: 'border-box' }` and report `entry.borderBoxSize[0]?.blockSize ?? entry.contentRect.height`. Cover with a test whose stub returns different content-box vs border-box heights and expect `resize(index, borderBox)`.

packages/0/src/maturity.json
```diff
@@ -719,8 +719,10 @@
       "category": "data"
     },
     "Virtualizer": {
-      "level": "draft",
-      "category": "data"
+      "level": "preview",
```
johnleider: **[suggestion]** `level` is `preview` (`since: null`, description filled) but the new-feature checklist items that `draft → preview` requires are missing: no `apps/docs/src/pages/components/data/virtualizer.md`, no `apps/docs/src/examples/components/virtualizer/basic.vue`, no row in `apps/docs/src/pages/components/index.md` Data table, no What's Included row in `packages/0/README.md` or root `README.md`. Every `@see` points at `https://0.vuetifyjs.com/components/data/virtualizer`, which 404s. `data-table.md` still footnotes Virtualizer as planned.

**Suggestion:** Add the docs page (category `data`, anatomy, basic `::: gn-example`, accessibility: tabindex on the scroller, consumer-supplied role/posinset), the basic example, index + README rows, and retarget the DataTable footnote. Keep `preview` only once that lands; otherwise leave `draft`.

packages/0/src/components/Virtualizer/VirtualizerRoot.vue
```diff
@@ -0,0 +1,179 @@
...
+    VirtualDirection,
+    VirtualItem,
+    VirtualState,
+  } from '#v0/composables/createVirtual'
+
+  export type VirtualizerRootContext = VirtualContext
+
+  export interface VirtualizerRootProps extends AtomProps {
+    /** The items to virtualize */
+    items?: readonly unknown[]
```
johnleider: **[suggestion]** `items?: readonly unknown[]` and `VirtualizerRootContext = VirtualContext` (no type param) leave slot `item.raw` as `unknown`. Sibling data/selection Roots generic the SFC (`DataGridRoot`, `ComboboxRoot`, `SelectRoot`, `TreeviewRoot`). The barrel example then has to write `(item.raw as typeof items[0]['raw']).name`, and that expression is also wrong: the slot prop shadows the outer `items`, so `items[0]` is a `VirtualItem`, not the source row.

**Suggestion:** `<script lang="ts" setup generic="T = unknown">`, type `items?: readonly T[]`, `createVirtual<T>(…)`, and `VirtualizerRootContext<T>`. The example becomes `{{ item.raw.name }}` (or a narrow cast only if T stays default-unknown).

packages/0/src/components/Virtualizer/VirtualizerRoot.vue
```diff
@@ -0,0 +1,179 @@
...
+
+  const virtual = createVirtual(_items, {
+    itemHeight,
+    height: _height,
+    overscan,
+    direction,
+    anchor,
+    anchorSmooth,
+    onStartReached,
+    onEndReached,
```
johnleider: **[suggestion]** `createVirtual` destructures options once and closes over `onStartReached` / `onEndReached` / `overscan` / `direction` / `anchor` / thresholds. Passing the prop identifiers in the factory argument list reads them at setup; a later `:on-end-reached` / `@end-reached` replacement (inline `() => load(page)` is the usual infinite-scroll form) never reaches `checkEdges`. Same snapshot for `overscan`/`direction` if those props change. Component props are expected to be live; the composable-only capture is not a good default once this is a public component.

**Suggestion:** Wrap callbacks so the latest prop is read at fire time (`onEndReached: d => onEndReached?.(d)`), and document that the rest of `VirtualOptions` is currently setup-time. Longer term, `createVirtual` should take `MaybeRefOrGetter` for those fields.

packages/0/src/components/Virtualizer/index.test.ts
```diff
@@ -0,0 +1,269 @@
...
+
+import { Virtualizer } from './index'
+
+// Utilities
+import { mount } from '@vue/test-utils'
+import { h, nextTick } from 'vue'
+
+// Types
+import type { VirtualizerRootContext } from './VirtualizerRoot.vue'
+
```
johnleider: **[suggestion]** Component tests live in `index.test.ts` (happy-dom) with a ResizeObserver stub that invents a fixed rect. testing.md requires `index.browser.test.ts` for components; Splitter/Pagination/Tabs follow that for layout. A virtualizer's contract is real scroll + real item measurement — the stub also reports the same 500px rect for the container and every Item, so the "visible window" test never sees `itemHeight: 40` layout. The a11y sweep fixture is not a substitute for functional browser coverage.

**Suggestion:** Add `index.browser.test.ts` that mounts a constrained-height Root with real items, asserts spacer heights and a visible count near `ceil(viewport/itemHeight)+2*overscan`, and scrolls the container to check the window moves. Keep the happy-dom file for context/ARIA passthrough.

packages/0/src/components/Virtualizer/VirtualizerItem.vue
```diff
@@ -0,0 +1,81 @@
...
+    index: number
+    /** Namespace for context injection from parent Virtualizer.Root */
+    namespace?: string
+  }
+</script>
+
+<script setup lang="ts">
+  defineOptions({ name: 'VirtualizerItem', inheritAttrs: false })
+
+  const attrs = useAttrs()
```
johnleider: **[suggestion]** Item's slot is `default: () => any` with no `VirtualizerItemSlotProps` and no `<slot v-bind>`. components.md requires `is<State>` + `attrs` on every compound sub-component. In renderless mode Atom has no element, so `useResizeObserver` never attaches and variable-height reporting is silently dead; the consumer also has no `attrs` (including `data-index`) to put on their host.

**Suggestion:** Export `VirtualizerItemSlotProps` with at least `{ index, attrs: { 'data-index': number } }` and bind them on the slot. In renderless, document that the consumer must put `attrs` on a real element (or skip Item and call `resize()` themselves).

packages/0/src/components/Virtualizer/VirtualizerRoot.vue
```diff
@@ -0,0 +1,179 @@
...
+    endThreshold,
+    momentum,
+    elastic,
+  })
+
+  const containerRef = useTemplateRef<AtomExpose>('container')
+  const el = toRef(() => toElement(containerRef.value?.element) ?? null)
+
+  watch(el, element => {
+    virtual.element.value = (element ?? undefined) as HTMLElement | undefined
```
johnleider: **[nit]** `virtual.element.value = (element ?? undefined) as HTMLElement | undefined` is the cast `components.md` tells authors not to use. `toElement` returns `Element | undefined`; `createVirtual.element` is typed `Ref<HTMLElement | undefined>`. Overflow/Image keep `Element | null` and cast at HTMLElement-only use sites.

**Suggestion:** Assign `toElement(...) as HTMLElement | undefined` only if you must match `VirtualContext.element`, or widen the composable's element type. Drop the `?? undefined` tautology either way.

PR #810, opened 2026-08-10
docs(Button): correct the loading grace default — 0, not 1 second

johnleider:
`ButtonRoot` defaults `grace = 0` (`ButtonRoot.vue:136`), so the loading indicator appears immediately unless a consumer passes `grace` — but three passages on the button page claimed a "built-in 1-second grace period". Rewritten to describe the opt-in `grace` prop with its real default.

Found during the Emerald docs section QA (PR #808) as an instance of the same docs-vs-source drift class that review was hunting.


johnleider:
Replaced by the dev-based PR — this branch's base was polluted with dev-lineage content (17 commits / 233 files instead of the single docs fix). The clean single-commit version is open against dev.

PR #811, opened 2026-08-10
docs(vapor): document the interop boundary rule

johnleider:
## What

Adds "The interop boundary rule" to the Vapor guide: `vaporInteropPlugin` overhead scales with the number of Vapor→vdom crossings, and compound components composed inline in Vapor templates multiply crossings (200 checkboxes ≈ 400 crossings). The rule: put the boundary above the repeated region — one classic wrapper — never inside it.

Includes a costly/cheap example pair, the in-repo gated bench numbers from #758 (+44% mount / +12% heap inline vs within-noise wrapped), and independent confirmation from a production Vue SPA (11k tests, cited anonymized with the consumer's permission): +66–78% mount / +49% heap inline at 1x/4x CPU throttle in real Chromium, everything collapsing to noise with one wrapper.

Also updates the interop limitations bullet to link the new section, and closes with the honest framing: the wrapper buys parity, not Vapor's update wins — vapor-native builds remain the actual fix.


johnleider:
Rebased onto current `master` (was 21 behind; `vapor.md` unchanged on master, no conflicts). Same two commits, same +47/−1 docs-only diff. Ready for review — content is the full interop-boundary section, not a stub.

Review focus: the costly/cheap pair, the #758 bench numbers vs the anonymized production confirmation, and whether the Vue distinct-regions footnote belongs here.

PR #812, opened 2026-08-10
docs(Button): correct the loading grace default — 0, not 1 second

johnleider:
Replaces #810 (whose branch base was polluted with dev-lineage content — 233 phantom files; closed). One commit, one file.

`ButtonRoot` defaults `grace = 0` (`ButtonRoot.vue:136`), so the loading indicator appears immediately unless a consumer passes `grace` — but three passages on the button page claimed a "built-in 1-second grace period". Rewritten to describe the opt-in `grace` prop with its real default.

Found during the Emerald docs section QA (PR #808) as an instance of the same docs-vs-source drift class that review was hunting.

### DIFF
--- a/apps/docs/src/pages/components/actions/button.md
+++ b/apps/docs/src/pages/components/actions/button.md
@@ -59,7 +59,7 @@ The Button component renders as a native `<button>` by default (or an anchor, ro
 
 ### Loading with Grace Period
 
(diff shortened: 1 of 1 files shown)


johnleider:
Superseded by the master-based PR #813 — John's call: it's a fix, so it rides the patch train; the original base-race is resolved (master's tip now verified).

PR #844, opened 2026-08-13
fix(EmSnackbar): announce error variants assertively instead of politely

### LINKED ISSUE #803: EmSnackbar: error variants announce politely — urgent is unreachable
Found during the extends-sweep (PR #802) while documenting withheld props.

Emerald's `EmSnackbar` withholds v0's `urgent` prop, and `variant` only sets `data-variant` (consumed exclusively by CSS). v0's `urgent` is what switches `role="alert"` vs `role="status"` and routes to the Portal announcer's assertive region — so **every Emerald snackbar, including `variant="error"`, renders `role="status"` and is announced politely**. A destructive-failure toast looks urgent and announces like a status update.

Two candidate fixes (both change public surface, hence out of the type-fidelity sweep):
1. Expose `urgent` (Pick it through) — consumer-controlled, matches v0.
2. Derive it from `variant` (`error` → assertive), optionally with `urgent` as an override.

Option 2 fits Emerald's variant-driven design language; option 1 is the smaller surface promise. The withholding comment added in #802 documents the current behavior either way.

sridhar-3009:
Fixes #803

## Root cause

`EmSnackbar` withholds v0's `urgent` prop entirely, and `variant` only sets `data-variant` (consumed by CSS). `urgent` is what switches `Snackbar.Root` between `role="alert"` (assertive, routed to the Portal's assertive live region) and `role="status"` (polite) — so every Emerald snackbar, including `variant="error"`, rendered `role="status"` and was announced politely by assistive tech. A destructive-failure toast looked urgent but announced like a routine status update.

## Fix

Went with the issue's option 2 (fits Emerald's variant-driven design language): `urgent` now derives from `variant` (`error` → assertive, everything else → polite), with an explicit `urgent` prop still available to override the derived default in either direction.

One non-obvious wrinkle I ran into and want to flag for review: a plain

```ts
const { urgent } = defineProps<{ urgent?: boolean }>()
```

does **not** give `urgent` the value `undefined` when the caller doesn't pass it — Vue's Boolean-prop casting resolves an absent Boolean-typed prop to `false`, not `undefined`, which permanently short-circuits `urgent ?? (variant === 'error')` (the `??` never sees `undefined`, since it's already `false`). The fix needs `urgent = undefined` explicitly in the destructure (verified empirically: this isn't a no-op — it's what tells Vue's runtime prop resolution to register an explicit `default: undefined` instead of falling through to the implicit Boolean-cast path). Left a comment on the line since it's easy to "simplify" back into the broken version.

## Test plan

- [x] Added `EmSnackbar.test.ts` (new — no test file existed) covering: `variant="error"` → `role="alert"`, `success`/`warning`/default → `role="status"`, and explicit `urgent` overriding the derived default in both directions
- [x] Reproduced the bug in isolation first (a debug harness showed `role="status"` even for `variant="error"` before the `= undefined` fix), confirmed the fix resolves it, then removed the scratch harness
- [x] `pnpm vitest run --project v0:unit --project emerald` → 4803 passed, 1 skipped
- [x] `pnpm --filter=@paper/emerald typecheck` and `pnpm eslint` clean on changed files
- [x] 100% branch/line coverage on the changed file
- [x] Added a changeset (patch bump for `@paper/emerald`)

### DIFF
--- a/.changeset/emsnackbar-urgent-role-803.md
+++ b/.changeset/emsnackbar-urgent-role-803.md
@@ -1,0 +1,8 @@
+---
+'@paper/emerald': patch
+---
+
+fix(EmSnackbar): announce error variants assertively instead of politely
+
+`variant="error"` now uses `role="alert"`. Other variants stay `role="status"`.
+Pass `urgent` to override either way.
--- a/packages/emerald/src/components/EmSnackbar/EmSnackbar.vue
+++ b/packages/emerald/src/components/EmSnackbar/EmSnackbar.vue
@@ -8,11 +8,10 @@
   export type EmSnackbarVariant = 'success' | 'error' | 'info' | 'warning' | 'neutral'
 
   /**
-   * Emerald withholds `urgent`, so every snackbar is `role="status"`. `variant` is
-   * purely visual (`data-variant`) and does not raise politeness — a `variant="error"`
-   * snackbar is still announced politely.
+   * `urgent` defaults to true for `variant="error"` and false otherwise;
+   * pass it explicitly to override the variant-derived default.
    */
-  export interface EmSnackbarProps extends Pick<SnackbarRootProps, 'id' | 'namespace'> {
+  export interface EmSnackbarProps extends Pick<SnackbarRootProps, 'id' | 'namespace' | 'urgent'> {
     variant?: EmSnackbarVariant
   }
 </script>
@@ -24,6 +23,10 @@
     id,
     namespace,
     variant = 'neutral',
+    // `= undefined` isn't a no-op here: without it, Vue's Boolean-prop casting
+    // resolves an absent `urgent` to `false` rather than `undefined`, which
+    // would permanently short-circuit the `??` fallback below.
+    urgent = undefined,
   } = defineProps<EmSnackbarProps>()
 </script>
 
@@ -33,6 +36,7 @@
     class="emerald-snackbar"
     :data-variant="variant"
     :namespace
+    :urgent="urgent ?? variant === 'error'"
   >
     <slot />
   </Snackbar.Root>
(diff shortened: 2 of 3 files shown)


.changeset/emsnackbar-urgent-role-803.md
```diff
@@ -0,0 +1,17 @@
+---
+'@paper/emerald': patch
+---
+
+fix(EmSnackbar): announce error variants assertively instead of politely
+
+`EmSnackbar` withheld v0's `urgent` prop entirely, and `variant` only drove
```
johnleider: **[suggestion]** Changeset body dumps the bug writeup and the Snackbar.Root / Portal announcer mechanism. The title already answers "does this affect me / must I do anything?" Releasing contract: never the mechanism.

**Suggestion:** Keep the title. Replace the body with the consumer delta: `variant="error"` now uses `role="alert"`. Other variants stay `role="status"`. Pass `urgent` to override either way.

packages/emerald/src/components/EmSnackbar/EmSnackbar.vue
```diff
@@ -7,13 +7,14 @@
...
-   */
-  export interface EmSnackbarProps extends Pick<SnackbarRootProps, 'id' | 'namespace'> {
+  export interface EmSnackbarProps extends Pick<SnackbarRootProps, 'id' | 'namespace' | 'urgent'> {
     variant?: EmSnackbarVariant
+    /**
+     * Announce assertively (role="alert") instead of politely (role="status").
+     * Defaults to true for `variant="error"` and false otherwise; pass explicitly
+     * to override the variant-derived default.
+     */
+    urgent?: boolean
```
johnleider: **[suggestion]** After `Pick<SnackbarRootProps, 'id' | 'namespace' | 'urgent'>`, redeclaring `urgent?: boolean` is a duplicate field. Types match so it is not a type error. The redeclare exists so Volar shows Emerald's variant-default JSDoc instead of v0's simpler line; it also freezes the type if v0 later widens `urgent`.

**Suggestion:** Optional. Drop the field and put the default on the interface JSDoc if you want the Pick surface tidy. Not blocking.

Issue #733, opened 2026-07-27
[Feature] usePopover: pluggable positioning adapter seam (bring your own engine)


johnleider:
## Problem

`usePopover` has exactly one positioning implementation and no way to supply another.

`packages/0/src/composables/usePopover/index.ts:125-132` emits CSS anchor positioning and nothing else:

```ts
const contentStyles = toRef(() => ({
  'position': 'fixed',
  'margin': 'unset',
  'inset-area': positionArea,
  'position-area': positionArea,
  'position-anchor': anchor,
  'position-try-fallbacks': positionTry,
}))
```

`grep -rl floating-ui packages/0/src` returns nothing. There is no JS positioning path in the package at all — not as a default, not as an option, not as a seam someone could fill.

### Who this affects

Everything that consumes `usePopover`:

- `packages/0/src/components/Popover/PopoverRoot.vue`, `PopoverContent.vue`, `PopoverActivator.vue`
- `packages/0/src/components/Tooltip/TooltipRoot.vue`
- `packages/0/src/components/Select/SelectRoot.vue`
- `packages/0/src/composables/createCombobox/index.ts`

When the CSS properties aren't understood they are silently ignored, so the content element renders at `position: fixed` with no offsets — pinned to the viewport origin, unanchored. Not degraded placement: no placement.

### Browser reality (verified against MDN browser-compat-data, not caniuse prose)

| Property | Chrome | Firefox | Safari |
|---|---|---|---|
| `anchor-name` | 125 | 147 | 26 |
| `position-try-fallbacks` | 128 (as `position-try-options` 125–127) | 147 | 26 |
| `position-area` | **129** (as `inset-area` 125–130) | **147** | **26** |

Anchor positioning reached Baseline "newly available" in **January 2026** — six months ago.

Two live consequences:

- **Firefox ESR 140** is the current ESR channel (`140.13.0esr` as of 2026-07-21; ESR 153 is only now rolling out as ESR-next). 140 < 147. Every enterprise and institutional deployment pinned to ESR gets unpositioned popovers, tooltips, selects, and comboboxes.
- **Safari 18.x** — the entire pre-Safari-26 line, still a material share of installed macOS/iOS — same.

Also worth noting: the emitted `inset-area` line covers Chromium 125–130 only; `inset-area` was **removed in Chrome 131**, so that key is dead weight on every current Chromium and load-bearing only for a narrow legacy band.

`apps/docs/src/pages/introduction/browser-support.md:135` already concedes the point:

[...]

PR #846, opened 2026-08-13
feat(usePopover): add a pluggable positioning-adapter seam

sridhar-3009:
Addresses #733

## Scope of this PR

This covers the checklist items that live in `packages/0`:

- [x] Define the abstract contract in `usePopover/adapters/adapter.ts`
- [x] `V0PopoverAdapter` reproducing current CSS anchor output byte-for-byte
- [x] Activator element registration (the missing handle) — `attachAnchor()`
- [x] Neutral placement descriptor + raw CSS passthrough
- [x] Thread the `adapter` option through `PopoverRoot`, `TooltipRoot`, `SelectRoot`, `createCombobox`
- [x] Docs page + a worked "bring your own engine" example that does **not** ship in the package
- [x] Snapshot-equivalent test proving the default path is unchanged (the full existing `usePopover` test suite passes unmodified against the new code path)

Non-goals from the issue are respected: no `floating-ui` dependency anywhere in `packages/0`, no bundled JS adapter, CSS anchor positioning stays the zero-cost default.

## Naming

Went with `PopoverAdapter` per the convention the issue itself cites (`.claude/rules/implementation.md`'s "composition noun matches the composable name verbatim"), over the more precise `PositionAdapter` the issue also raised as worth settling in review. Easy to rename before merge if you'd rather have `PositionAdapter` — it's a single identifier used consistently, no structural difference either way.

## Design notes

**Context shape.** `PopoverAdapterContext` carries `anchorName` (the CSS custom-ident), `anchorEl`/`contentEl` (both element refs — the concrete blocker the issue called out, since `usePopover` previously only ever saw the content element), `isOpen`, `placement`, and `positionTry`. `setup(context)` returns the reactive style object for the content element — for `V0PopoverAdapter` that's the existing six CSS declarations; for a JS engine it'd be computed `top`/`left`/`position`.

**Placement descriptor.** `derivePlacement()` (`adapters/placement.ts`) does a best-effort parse of the `positionArea` CSS value into `{ side, align }` — covers the common cases (`'bottom'`, `'top span-left'`, logical `block-start`/`block-end`) and falls back to `bottom`/`center` for anything it doesn't recognize, with `raw` always carrying the original value verbatim. I didn't try to build a complete grammar for every `position-area` compound value — the issue asked for a "raw passthrough escape hatch so the CSS path loses no expressiveness," which `raw` provides regardless of what the normalized descriptor captures.

**`createCombobox`'s option is `positionAdapter`, not `adapter`** — it already has an `adapter?: ComboboxAdapter` option for query filtering, so reusing the name would be ambiguous. `Popover.Root`/`Tooltip.Root`/`Select.Root` don't have that collision and use plain `adapter`.

**Lifecycle.** `adapter.dispose?.()` runs on `onScopeDispose`, with Vue 3.5's `failSilently` flag — `usePopover()` is routinely called outside an active effect scope (directly in tests, for instance), which would otherwise emit a spurious warning.

## Test plan

- [x] The full pre-existing `usePopover` test suite (17 tests) passes unchanged against the new adapter-based `contentStyles` — this is the "default output identical to today" proof, since those tests assert exact CSS output
- [x] Added unit tests for `derivePlacement()`, `V0PopoverAdapter`, and the `usePopover` adapter seam (custom adapter receiving the right context, `attachAnchor`/`attach` populating `anchorEl`/`contentEl`, dispose lifecycle) — 100% line/branch coverage on all new files
- [x] Added two browser tests to `Popover/index.browser.test.ts` mounting the full `Root`+`Activator`+`Content` stack with a custom adapter, proving both DOM elements reach it and its styles are applied to (not merged with) the content element
- [x] `pnpm vitest run --project v0:unit --project v0:browser` → 6561 passed, 3 skipped (full suite, no regressions)
- [x] `pnpm --filter=@vuetify/v0 typecheck` clean
- [x] `pnpm eslint` clean on all changed files
- [x] `pnpm vitest run packages/0/src/surface.test.ts` — updated the frozen public-surface list for the 3 new exports (`PopoverAdapter`, `V0PopoverAdapter`, `derivePlacement`)
- [x] Added a changeset (minor bump for `@vuetify/v0`)

One environment caveat: I couldn't get a fully clean local `pnpm run build:docs` — this checkout has pre-existing case-collision issues on macOS's case-insensitive filesystem between intentionally-paired files like `Equalizer.vue`/`equalizer.vue` (unrelated to this PR, reproducible on a pristine `dev` checkout with zero changes). I reviewed the `use-popover.md` diff manually for syntax correctness against the page's existing conventions; CI's `docs-check` runs on a case-sensitive runner and will give the real signal.


packages/0/src/composables/usePopover/adapters/placement.ts
```diff
@@ -0,0 +1,37 @@
...
+  const tokens = positionArea.trim().split(/\s+/)
+  /* v8 ignore next -- defensive: split() on a string always yields at least one element */
+  const first = tokens[0] ?? 'bottom'
+
+  let side: PopoverSide = 'bottom'
+  if (SIDES.has(first)) {
+    side = first as PopoverSide
+  } else if (first === 'block-start' || first === 'inline-start') {
+    side = 'top'
+  } else if (first === 'block-end' || first === 'inline-end') {
```
johnleider: inline-start / inline-end are mapped onto the block axis (	op / ottom).

Default V0PopoverAdapter is fine — it uses 
aw. A JS engine consuming side / lign gets the wrong axis.

inline-start → left, inline-end → 
ight (keep lock-start → 	op, lock-end → ottom). Add cases to placement.test.ts.

johnleider (commented):
Architecture skim: the adapter seam is right (PopoverAdapter + V0PopoverAdapter CSS default, no floating-ui, attachAnchor, positionAdapter on combobox). One descriptor bug before merge — inline on placement.ts.

sridhar-3009:
Good catch, fixed in b410039 — `inline-start`/`inline-end` now map to `left`/`right` instead of the block axis. Added a test case asserting the corrected axis, and swapped the if/else-if chain for a keyword lookup table since the switch was flagged by `unicorn/prefer-switch`. Full suite still green, coverage still 100% on the changed files.

PR #861, opened 2026-08-16
docs: make custom themes selectable

J-Sek:
Problem:
1. open docs theme customization
2. configure custom theme
3. try to select (not working)

Note:
- when starting from a11y themes (that have tinted scrollbars) the custom theme does not allow changing scrollbar thumb variable, so I included quick fix for this as well


johnleider:
Superseded by the reimplementation on the #871 controller architecture — see the new PR. The original toggle logic predates createThemeToggle and no longer applies; the storage-key fix and scrollbar/color-input commits were ported.

PR #862, opened 2026-08-16
feat(toHighlight): add ignore-accents support

J-Sek:
blocks [vuetify#22922](https://github.com/vuetifyjs/vuetify/pull/22922)

### DIFF
--- a/.changeset/highlight-ignore-accents.md
+++ b/.changeset/highlight-ignore-accents.md
@@ -4,6 +4,6 @@
 
 feat(toHighlight): add `ignoreAccents` so a plain query matches accented text
 
-`toHighlight(text, query, { ignoreAccents })` folds diacritics before matching and maps the ranges back onto the source, so `zurich` highlights *Zürich* with its umlaut intact. It is directional — `'target'` folds only the text, `'query'` only the search term, `true` both sides — and covers the letters NFD leaves alone (`ł`, `ø`, `ß`, `æ`, …).
+`toHighlight(text, query, { ignoreCase: true, ignoreAccents: true })` folds diacritics before matching and maps the ranges back onto the source, so `zurich` highlights *Zürich* with its umlaut intact. It is directional — `'target'` folds only the text, `'query'` only the search term, `true` both sides — and covers the letters NFD leaves alone (`ł`, `ø`, `ß`, `æ`, …).
 
 The matcher behind it ships as `findMatchRanges(text, query, { ignoreCase, ignoreAccents, matchAll })` for filters that need the same ranges without the chunking.
--- a/.changeset/highlight-ignore-accents.md
+++ b/.changeset/highlight-ignore-accents.md
@@ -4,7 +4,7 @@
 
 feat(toHighlight): add `ignoreAccents` so a plain query matches accented text
 
-`toHighlight(text, query, { ignoreCase: true, ignoreAccents: true })` folds diacritics before matching and maps the ranges back onto the source, so `zurich` highlights *Zürich* with its umlaut intact. It is directional — `'target'` folds only the text, `'query'` only the search term, `true` both sides — and covers the letters NFD leaves alone (`ł`, `ø`, `ß`, `æ`, …).
+`toHighlight(text, query, { ignoreCase: true, ignoreAccents: true })` folds diacritics before matching and maps the ranges back onto the source, so `zurich` highlights *Zürich* with its umlaut intact. It is directional — `'target'` folds only the text, `'query'` only the search term, `true` both sides — and covers common letters NFD leaves alone (`ł`, `ø`, `ß`, `æ`, …).
 
 With `ignoreCase`, Greek final sigma folds to `σ` so `ΣΟΦΟΣ` and `σοφος` match each other.
 
--- a/.changeset/highlight-ignore-accents.md
+++ b/.changeset/highlight-ignore-accents.md
@@ -1,0 +1,9 @@
+---
+"@vuetify/v0": minor
+---
+
+feat(toHighlight): add `ignoreAccents` so a plain query matches accented text
+
+`toHighlight(text, query, { ignoreAccents })` folds diacritics before matching and maps the ranges back onto the source, so `zurich` highlights *Zürich* with its umlaut intact. It is directional — `'target'` folds only the text, `'query'` only the search term, `true` both sides — and covers the letters NFD leaves alone (`ł`, `ø`, `ß`, `æ`, …).
+
+The matcher behind it ships as `findMatchRanges(text, query, { ignoreCase, ignoreAccents, matchAll })` for filters that need the same ranges without the chunking.
--- a/.changeset/highlight-ignore-accents.md
+++ b/.changeset/highlight-ignore-accents.md
@@ -6,4 +6,6 @@ feat(toHighlight): add `ignoreAccents` so a plain query matches accented text
 
 `toHighlight(text, query, { ignoreCase: true, ignoreAccents: true })` folds diacritics before matching and maps the ranges back onto the source, so `zurich` highlights *Zürich* with its umlaut intact. It is directional — `'target'` folds only the text, `'query'` only the search term, `true` both sides — and covers the letters NFD leaves alone (`ł`, `ø`, `ß`, `æ`, …).
 
+With `ignoreCase`, Greek final sigma folds to `σ` so `ΣΟΦΟΣ` and `σοφος` match each other.
+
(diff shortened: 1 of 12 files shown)


PAJO2018 (approved):

packages/0/src/utilities/diacritics.ts
```diff
@@ -0,0 +1,166 @@
...
+
+  if (ignoreCase) {
+    needle = needle.toLocaleLowerCase()
+  }
+
+  if (needle.length === 0) {
+    return []
+  }
+
+  if (!foldTarget) {
```
johnleider: **[bug]** `ignoreCase` without folding now highlights the wrong slice.

Master `toHighlight` used `index + term.length`. This path `toLocaleLowerCase()`s the haystack and returns `collect` offsets with **no** source map, using `needle.length` (the lowercased query).

```
'İ'.toLocaleLowerCase() === 'i\u0307'   // length 2
toHighlight('İstanbul', 'İ', { ignoreCase: true })
  master:  [0, 1] → 'İ'
  this PR: [0, 2] → 'İs'
```

Matches *after* an expanding prefix (`'stan'` in `'İstanbul'`) were already wrong on master (`slice(2, 6) === 'tanb'`). Same fix covers both.

**Suggestion:** run `foldWithMap` whenever `ignoreCase || foldTarget` (lowercase-only when not folding accents). Add `'İstanbul'` / `'İ'` / `ignoreCase: true` with accents off as a regression test.

packages/0/src/utilities/diacritics.ts
```diff
@@ -0,0 +1,166 @@
...
+    const haystack = ignoreCase ? text.toLocaleLowerCase() : text
+
+    return collect(haystack, needle, matchAll)
+  }
+
+  const { folded, map } = foldWithMap(text, ignoreCase)
+
+  return collect(folded, needle, matchAll)
+    .map(([start, end]) => [map[start]!, map[end]!])
```
johnleider: **[bug]** Half of a multi-char fold emits a zero-width range.

`ß→ss` / `þ→th` / `æ→ae` write the same source index twice. A query that consumes one output unit maps to `[i, i]`:

```ts
findMatchRanges('ß', 's', { ignoreAccents: true })  // [[0, 0]]
findMatchRanges('straße', 's', { ignoreAccents: true, matchAll: true })
// [[0, 1], [4, 4], [4, 5]]
```

`toHighlight.merge` drops `start >= end`, so the match vanishes. Direct `findMatchRanges` consumers (the changeset sells this to filters) get an empty slice. Tests only cover the full fold (`strasse` → `straße`).

**Suggestion:** if a mapped match touches a source character, span that whole character; never emit `start === end`. Lock the contract (`[]` vs expand to `[[0, 1]]`) with a test.

.changeset/highlight-ignore-accents.md
```diff
@@ -0,0 +1,9 @@
+---
+"@vuetify/v0": minor
+---
+
+feat(toHighlight): add `ignoreAccents` so a plain query matches accented text
+
+`toHighlight(text, query, { ignoreAccents })` folds diacritics before matching and maps the ranges back onto the source, so `zurich` highlights *Zürich* with its umlaut intact. It is directional — `'target'` folds only the text, `'query'` only the search term, `true` both sides — and covers the letters NFD leaves alone (`ł`, `ø`, `ß`, `æ`, …).
```
johnleider: **[suggestion]** `fold()` does not lowercase. `toHighlight('Zürich', 'zurich', { ignoreAccents: true })` is a miss.

The JSDoc / `utilities.md` samples already pass `{ ignoreCase: true, ignoreAccents: true }`. This sentence and the Recipes lead do not.

**Suggestion:** pair the two flags everywhere `zurich` / *Zürich* is the motivating example.

apps/docs/src/pages/composables/transformers/to-highlight.md
```diff
@@ -130,6 +130,22 @@ before chunking, so unordered or overlapping input is normalized for you.
 
 :::
 
+## Recipes
+
+### Accent-insensitive search
+
+`ignoreAccents` folds diacritics before matching, then maps the ranges back onto the source
+string — the rendered chunks keep their original characters. It is directional: `'target'` folds
+only the text so a plain `zurich` reaches *Zürich*, `'query'` folds only the search term so a
```
johnleider: **[suggestion]** Two claims here do not match the implementation or the demo under it.

1. A plain `zurich` does not reach *Zürich* without `ignoreCase: true` — folding is not case-folding.
2. A pasted `Trø` does not reach plain `Tromso` in the example — the list has `Tromsø`, and pasting the full `Tromsø` in `'query'` mode folds to `Tromso`, which is not a substring of `Tromsø`. `Trø` only matches as a prefix that dies before `ø`.

**Suggestion:** pair `ignoreCase` in the `zurich` sentence; point the `'query'` sentence at a string the demo actually contains (or add a plain `Tromso` / `cafe` row).

apps/docs/src/examples/composables/to-highlight/accents.vue
```diff
@@ -0,0 +1,78 @@
...
+  import { computed, shallowRef } from 'vue'
+  import type { IgnoreAccents } from '@vuetify/v0'
+
+  const query = shallowRef('zurich')
+  const mode = shallowRef<'off' | 'target' | 'query' | 'both'>('target')
+
+  const hints = {
+    off: 'The query must match the exact characters.',
+    target: 'Accents in the text are folded — type “zurich” or “krakow”.',
+    query: 'Accents in the query are folded — copy “Trø” to match plain text.',
```
johnleider: **[suggestion]** This hint is false for the list below.

There is no plain `Tromso`. Pasting `Tromsø` in `'query'` mode folds to `Tromso` and misses `Tromsø`. `Trø` only matches because it is a prefix that ends before `ø`.

Also: `off` says "exact characters" but the example hardcodes `ignoreCase: true`, so off is accent-sensitive, not exact. The `ignoreAccents` mapper is a cheap derivation — `toRef`, not `computed` (`MessageRow.vue` is the sibling precedent).

**Suggestion:** add a plain `Tromso` / `cafe` row, or change the hint to an action that works (`paste “café” to match “cafe”`).

userquin:
We are using toLocaleLowerCase at diacritics, we may have some issue without using i18n lang... that method will use browser/navigator locale.

For example, having en-US at browser and using some german word.

I'm just asking...

johnleider:
@userquin The API concern is real; the German example isn't.

`toLocaleLowerCase()` with no argument uses the **host** locale (browser / Node `LANG`). That's a problem for a pure utility: same call, different machine, different ranges. SSR is the sharp version — Node locale vs client locale can disagree, so highlighted markup hydrates wrong.

`en-US` and `de` lowercase `Straße` / `STRASSE` / `ß` identically. The actual landmine is Turkic I:

```ts
'I'.toLowerCase()                   // 'i'          always
'I'.toLocaleLowerCase('tr')         // 'ı'          dotless
'İstanbul'.toLowerCase()            // 'i̇stanbul'  i + combining dot
'İstanbul'.toLocaleLowerCase('tr')  // 'istanbul'
```

A `tr` browser and an `en-US` browser will not highlight the same `I` / `İ` / `i` pairs.

This PR inherited `toLocaleLowerCase` from master `toHighlight`. `createFilter` already uses locale-independent `toLowerCase()`. The new public `findMatchRanges` should match that, not the host.

Don't thread `useLocale()` in here. `findMatchRanges` is `#v0/utilities` — no Vue, no plugin. If we ever want German/Turkish-correct case, add an explicit `locale?: string` later. Silent navigator language is the worst option.

I'll switch both `toLocaleLowerCase()` calls to `toLowerCase()`. The `İ` source map still matters; Unicode default still expands `İ` to two units.

userquin:
Maybe we can use `'İstanbul'.normalize('NFD').toLowerCase()` (this shouldn't break)

J-Sek:
```ts
'ΣΟΦΟΣ'.toLowerCase() // 'σοφος'
'σοφος'.toUpperCase() // 'ΣΟΦΟΣ'
```

And apparently it is correct, lol

userquin:
> ```ts
> 'ΣΟΦΟΣ'.toLowerCase() // 'σοφος'
> 'σοφος'.toUpperCase() // 'ΣΟΦΟΣ'
> ```
> 
> And apparently it is correct, lol

it looks like `toLowerCase/toUpperCase` is doing its work, there is no need to use `normalize('NFD')`

johnleider:
Looking for final comments.

PR #891, opened 2026-08-19
fix(EmKanban): announce rejected drops instead of leaving the live region stale

johnleider:
From the kanban dogfood memo's verified defects: `kanban.transfer` returns `undefined` for eight gated cases and `onDrop` bailed silently — a rejected keyboard drop left the live region holding the stale pickup message, so a screen-reader user got no feedback at all. The rejection now announces ('Move not allowed, the card stayed where it was.', mirroring the existing cancel wording); same-column no-op drops can't false-positive since `sortable.move` returns the ticket even when the index is unchanged.

Changeset parked in `packages/emerald/changesets-parked/` per the private-package convention.

Verified: emerald typecheck + build green; the announcement chain is source-verified (onDrop → context.announce → message ref → `role=status` region). A live browser repro is currently impossible by construction — no shipped example can reach a rejection because per-column `disabled`/`accept` aren't surfaced as DS props (already a listed gap in the memo) and board-level disabled gates pickup before any drop. That gap is exactly why this went unnoticed.

### DIFF
--- a/packages/emerald/changesets-parked/emkanban-reject-announce.md
+++ b/packages/emerald/changesets-parked/emkanban-reject-announce.md
@@ -1,0 +1,9 @@
+---
+'@paper/emerald': patch
+---
+
+fix(EmKanban): announce rejected drops instead of leaving the live region stale
+
+A drop gated by `disabled`, `accept`, or an unknown id used to bail silently,
+so keyboard and screen-reader users still heard the pickup message. The board
+now announces "Move not allowed, the card stayed where it was."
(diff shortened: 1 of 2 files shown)


johnleider:
Superseded by #894.

This PR targeted `dev`; the change is a patch-level a11y fix and belongs on `master` (same class as #844). Flipping the base here would have dragged 14 unrelated `dev` commits onto master, so `f86d3d2` was cherry-picked onto a master-based branch instead. #894 also adds the happy-dom live-region test and tightens the comment/changeset copy from review.

johnleider:
Closing in favor of #894 (master-based cherry-pick).

PR #892, opened 2026-08-19
fix(useFeatures): persist a delta of user-touched flags, not the full selection

johnleider:
Fix-forward for the persist option merged in #388, per the design review against the theme-persist doctrine (persist the user's explicit choice; stored state must never override future defaults).

**Was:** full `selectedIds` snapshot + restore reconciling every flag with select/unselect — a flag newly enabled-by-default would be unselected forever for returning users, and adapter-registered flags missed restore (it runs before setup).

**Now:** `{ enabled, disabled }` delta of user-touched flags relative to registration defaults. Untouched flags follow code/adapter defaults across releases; saved overrides apply on late registration (runtime *and* adapter — tested); unregistered entries prune on the next write; toggle-back-to-default drops the entry; the old array format is ignored by the shape guard (acts as migration, though dev's format never shipped anywhere).

Intent tracking: a `touched` set (shallowReactive — the persist watch needs a reactive dep even while empty) marked by the context's select/unselect/toggle; plumbing reaches the plugin hooks via the same WeakMap pattern the previous implementation used, nothing added to the public surface. Documented limitation: ticket self-methods and bulk ops bypass intent tracking — no registry-layer selection events exist to hook.

The #388 changeset is edited in place (dev hasn't released), so the eventual minor gets one clean changelog entry. Docs Persistence section + FAQ updated.

useFeatures 105/105; full v0:unit sweep 4,762 green; typecheck clean.


johnleider:
Coverage bounce addressed (+9 tests, 114/114 focused, 4,771 full sweep): the intent-tracking limitation is now pinned as contract rather than prose, reset()/replay-ordering/storage-transition/numeric-id edges covered, and UNSAFE_KEYS ruled inapplicable (array delta, Map application — no object-key sink).

Two findings out of writing them:
- **Real bug fixed** (one line): a `{ value: true, disabled: true }` flag never auto-selects, but the delta baseline recorded `isEnabled` — a no-op toggle persisted a phantom `disabled` entry for a change the user never made. Baseline now reads the registry state the flag actually boots with; differentially verified (reverting the line fails exactly the new test).
- **Behavioral nuance, pinned as-is, your call whether it deserves its own issue:** `reset()` does not restore registration defaults in-session — `model.reset()` clears the selection outright, so even default-enabled flags end up unselected until next boot. The test pins the real contract (storage → null; defaults return on a fresh boot).

johnleider:
Reset override added per review: `reset()` now restores registration defaults in-session (replays the baseline map), so in-session and post-reload state agree — and it drops un-applied pending restore entries, closing a real resurrection path the new test caught: a pre-reset override whose flag never registered this session lived only in `pending`, changed nothing reactive when cleared, and the persist watch never wrote — storage kept it for the next boot. Fix: `pending` is a tracked shallowRef and the delta always yields a fresh object (the stored no-overrides value is now the empty delta; legacy `null` still skipped both directions).

Disabled-flag decision: replay respects the guard — a flag disabled after boot stays unselected on reset, consistent with the standing disabled-mutation ruling and with the restore path.

116/116 focused, 4,773 sweep, typecheck clean.

Issue #899, opened 2026-08-20
[Feature] createPopoverPlugin: app-wide positioning-adapter default, plus adapter option on createTooltipPlugin


johnleider:
Decided follow-up to the #846 adapter seam (refs #733, pairs with #897).

Two additions, one precedence chain:

1. **`createPopoverPlugin({ adapter })`** — new plugin providing the app-wide positioning-adapter default. `usePopover` consults it via optional injection (Treeview null-fallback pattern); no plugin installed → `V0PopoverAdapter`, zero-config behavior unchanged. One line switches every popover-derived surface (Popover, Tooltip, Select, Combobox):
   `app.use(createPopoverPlugin({ adapter: new FloatingUIPopoverAdapter() }))`
   Also becomes the home for future popover-wide defaults (e.g. `positionTry`).

2. **`adapter` option on `createTooltipPlugin`** — a tooltip-scoped override so all tooltips can run a different engine/middleware than the rest of the app's popovers.

**Precedence (explicit, most-specific wins):** per-instance `adapter` option → tooltip plugin `adapter` (tooltip surfaces only) → popover plugin `adapter` → `V0PopoverAdapter`.

Sequencing: land with or after #897 (the first-party floating-ui adapter) on `dev`.

johnleider:
Landed in https://github.com/vuetifyjs/0/pull/917 (merged to `dev`; GitHub won't auto-close from a non-default base). #916 was the same plugin work — closed as duplicate.

Issue #897, opened 2026-08-20
[Feature] usePopover: ship a first-party floating-ui adapter


johnleider:
Follow-up to #846 (refs #733). The adapter seam lands there with CSS anchor positioning as the default; a first-party floating-ui adapter should ship from v0 itself rather than being left to consumers.

Shape (per the flagsmith/launchdarkly/posthog precedent in `features/adapters/*`):
- Subpath export (e.g. `@vuetify/v0/popover/adapters/floating-ui`) with `@floating-ui/dom` as an optional peerDependency
- `computePosition` + `autoUpdate`, default `offset`/`flip`/`shift` middleware, constructor accepts middleware overrides
- `positionTry` documented as ignored under this adapter (`flip()` covers the intent)
- Unit tests (mocked) + one browser test asserting real coordinates
- Docs: flip the use-popover FAQ to point at the shipped adapter; keep the bring-your-own sketch as an appendix

Lands on `dev` after #846 merges.

johnleider:
Landed in https://github.com/vuetifyjs/0/pull/917 (merged to `dev`; GitHub won't auto-close from a non-default base).

PR #916, opened 2026-08-24
feat(usePopover): createPopoverPlugin app-wide adapter default + tooltip adapter option

johnleider:
Decided follow-up to the #846 adapter seam (refs #733, pairs with #897).

The seam made the positioning engine pluggable, but only per instance — switching every popover-derived surface (Popover, Tooltip, Select, Combobox) onto a JS engine meant passing `adapter` at every call site. This adds the two configuration levels from the issue:

- **`createPopoverPlugin({ adapter })`** — app-wide positioning-adapter default. `usePopover` consults it via optional injection (private consumer, `{ adapter: null }` fallback), so no plugin installed still means `V0PopoverAdapter` and unchanged zero-config output. Provided under `v0:popover:defaults`, keeping the Popover component context's `v0:popover` key untouched. Also the future home for popover-wide defaults (e.g. `positionTry`).
- **`adapter` option on `createTooltipPlugin`** — tooltip-scoped override, exposed as `TooltipContext.adapter` and threaded through `Tooltip.Root`, so all tooltips can run a different engine/middleware than the rest of the app's popovers.

**Precedence (explicit, most-specific wins):** per-instance `adapter` option → tooltip plugin `adapter` (tooltip surfaces only) → popover plugin `adapter` → `V0PopoverAdapter`.

Also included:

- `usePopover/index.ssr.test.ts` — required by the plugin SSR coverage meta-test now that `usePopover` hosts a plugin trinity
- docs: `use-popover` "App-wide adapter default" section, `use-tooltip` adapter row, `plugins` guide entry
- `docs(build)`: the api-whitelist generator now picks up named export lists (`export { a, b }`), which also fixed a pre-existing logger-trinity misattribution
- changeset (`@vuetify/v0`: minor)

Sequencing per the issue: no code dependency on #897 (tests use stub adapters) — coordinate merge order only if #897 lands first and touches the same lines.

closes https://github.com/vuetifyjs/0/issues/899


johnleider:
Closing as duplicate of #917.

Both PRs implement `createPopoverPlugin` + the tooltip `adapter` option (#899) on the same files (`usePopover/index.ts`, `useTooltip`, Tooltip.Root, surface, the same docs pages). #917 also ships the first-party floating-ui adapter (#897) and the inspect follow-ups (Content placement, Combobox `positionAdapter`). Merging both would conflict.

#917 is merged. #899 closes there.

PR #921, opened 2026-08-26
chore(play): mark first publish public so OIDC can create the package

johnleider:
Release #895 published `@vuetify/v0@1.1.0` then died on `@vuetify/play@0.1.0`:

`404 Not Found - PUT https://registry.npmjs.org/@vuetify%2fplay`

New scoped packages default to restricted; a 404 is what npm returns when the package does not exist yet. `publishConfig.access: public` lets the next master Release run create it.

`v1.1.0` GitHub release is already minted. This unblocks the play publish + `@vuetify/play@0.1.0` GitHub release on the recovery run (v0 is skipped as already on npm).

PR #922, opened 2026-08-26
chore: retract 1.1.0 so DataTable/DataGrid can own the minor

johnleider:
\`@vuetify/v0@1.1.0\` already published from #895 (Splitter + play hash + kanban) without DataTable/DataGrid, which are still on \`dev\`.

This:
1. Reverts \`chore: version packages (#895)\` — \`packages/0\` back to 1.0.5, pending changesets restored
2. Marks \`@vuetify/play\` \`publishConfig.access: public\` for the first OIDC create
3. Adds a one-shot \`Unpublish v0 1.1.0\` workflow (OIDC). Local \`npm unpublish\` 404s; the token that published it is GHA.

After merge: run **Actions → Unpublish v0 1.1.0 → Run workflow**, then delete that workflow. Merge \`dev\` into \`master\` for the real 1.1.0 cut.

Supersedes #921.


johnleider:
Cannot unpublish @vuetify/v0@1.1.0 — npm blocks it because @paper/genesis@1.0.2 depends on ^1.0.5. Reverting master to 1.0.5 would fight the next Version Packages PR. 1.1.0 stays; DataTable/DataGrid ship as 1.2.0.

